PR #4, opened 2016-03-09
Use WAR/Tomcat, support plugins, document setting memory

### LINKED ISSUE #3: Using Tomcat base container but running self-contained Geoserver bin?
I'm confused as to why you're starting with the Tomcat 8.0 base container but are opting to instead run the self-contained Geoserver binary which runs on Jetty 6.1.8.

My use case is to run this container as part of an integration test and connect the container to a database container and I'd like to use JNDI to do so, but obviously writing the JNDI connection info into /usr/local/tomcat/conf/context.xml doesn't work and downloading the postgres driver to /usr/local/tomcat/lib isn't going to work.

Why not just run the Geoserver war on the Tomcat container?

shane-axiom:
This commit fixes #3 by using the GeoServer WAR download with Tomcat instead of the self-contained Jetty version. It also adds support for building the image with GeoServer plugins and adds instructions on setting Tomcat properties such as maximum memory.


shane-axiom:
If accepted this will need to be applied to the other branches.

Dockerfile
```diff
@@ -16,19 +16,32 @@ RUN apt-get -y update
...
-    unzip /tmp/resources/geoserver.zip -d /opt && mv -v /opt/geoserver* /opt/geoserver
-ENV GEOSERVER_HOME /opt/geoserver
-ENV JAVA_HOME /usr/
-
-#ENTRYPOINT "/opt/geoserver/bin/startup.sh"
-CMD "/opt/geoserver/bin/startup.sh"
-EXPOSE 8080
+    unzip /tmp/resources/geoserver.zip -d /tmp/geoserver \
+    && mv /tmp/geoserver/geoserver.war /usr/local/tomcat/webapps/geoserver.war \
+    && unzip /usr/local/tomcat/webapps/geoserver.war -d /usr/local/tomcat/webapps/geoserver \
```
isuftin: This part should be done by Tomcat when Tomcat first starts up. Tomcat should explode the war into its own directory. This step is probably not necessary.

README.md
```diff
@@ -76,10 +83,23 @@ mkdir -p ~/geoserver_data
...
+You need to ensure the ``geoserver_data`` directory has sufficient permissions
 for the docker process to read / write it.
 
+## Setting Tomcat properties
 
+To set Tomcat properties such as maximum heap memory size, create a `setenv.sh` file such as:
+
+```shell
+JAVA_OPTS="$JAVA_OPTS -Xmx1536M -XX:MaxPermSize=756M"
+JAVA_OPTS="$JAVA_OPTS -Djava.awt.headless=true -XX:+UseConcMarkSweepGC -XX:+CMSClassUnloadingEnabled"
```
isuftin: One thing you may want to mention is creating a way to set the Geoserver data directory. 

http://docs.geoserver.org/stable/en/user/datadirectory/data-dir-setting.html#java-system-property

You may want to add that ability since it allows a user of this container to persist their Geoserver data.

Take a look at what I've done here:
- https://github.com/USGS-CIDA/DSASweb/blob/master/geoserver/dsas-geoserver/src/main/resources/docker/Dockerfile
- https://github.com/USGS-CIDA/DSASweb/blob/master/geoserver/dsas-geoserver/src/main/resources/docker/setenv.sh

isuftin:
:clap: This is great!

Dockerfile
shane-axiom: The manual war unzipping is done here so that jars from the plugin zips can be added to GeoServer's `WEB-INF/lib` directory. Otherwise the GeoServer WAR would have to be unzipped to a temp directory, the plugin jars added there, and the WAR rezipped only to be exploded by Tomcat again. Seemed better to just do it this way.

README.md
shane-axiom: @isuftin The current [README](https://github.com/kartoza/docker-geoserver#storing-data-on-the-host-rather-than-the-container) covers how to make the data dir persistent via a Docker volume, so I just set the GEOSERVER_DATA_DIR environment var in the Dockerfile to the already mentioned path for consistency (`/opt/geoserver/data_dir`). This approach is similar to other images I've seen (e.g. Postgres).

shane-axiom:
@isuftin Thanks for checking it over!

Please don't merge this yet, working on some revisions with @alexgleith.

timlinux:
Wonderful to get your patch! Just give me a nod when you are ready to merge. Thanks, Tim.

Dockerfile
alexgleith: I agree with shane, You do need to manually unzip GeoServer, as he says, to install plugins. And potentially install a modified web.xml.

shane-axiom:
@timlinux Ok, I think this is finally ready for review. @alexgleith added installation of the [JAI/ImageIO extensions](http://docs.geoserver.org/stable/en/user/production/java.html#install-native-jai-and-jai-image-i-o-extensions) which will greatly improve performance. He also pointed out that the Oracle JDK is recommended by GeoServer, so I added a build-arg to optionally replace the OpenJDK with OracleJDK during the build (note that the stock image still uses OpenJDK, so there won't be any licensing issues with putting it on Docker Hub). In addition to adding support for building with plugins, I also added the ability to include static file system overlays during the build (useful for including static data directories, extra jars, and other advanced uses). There's also a build arg (`TOMCAT_EXTRAS`) for removing the default Tomcat documentation, examples, and manager webapp. Finally, we cleaned up some scripts that were either unused or contained invalid configurations.

All this work and collaboration has caused a bit of commit sprawl. If that's an issue I can check into squashing some of it.

Please let me know if you have any questions!

shane-axiom:
Oh, one more thing, it supports installing the JCE policy files for the Oracle JDK during the build if they're included at `resources/jce_policy.zip`.

README.md
isuftin: @shane-axiom No, I think that's fine. 

Great PR!

timlinux:
@shane-axiom could I ask you to add yourself to the README.md in the credits section at the end too as part of this patch?

shane-axiom:
@timlinux Updated, I added @alexgleith also.

timlinux:
Thanks @shane-axiom !

shane-axiom:
@timlinux Heh, looks like I botched the spacing on the [README credits](https://github.com/kartoza/docker-geoserver/blob/master/README.md#credits), do you mind fixing?

Also, will you handle merging this with the various version branches, or should I make PRs for those?

Thanks!

timlinux:
> Heh, looks like I botched the spacing on the README credits, do you mind fixing?

@shane-axiom no worries - I already fixed it.

> Also, will you handle merging this with the various version branches, or should I make PRs for those?

@shane-axiom I can do it but I probably won't have a chance to do it before the weekend. The master branch built fine in the docker hub [1] (great!) so we can merge these changes across to the other branches that are build on docker - currently:

screen shot 2016-03-15 at 22 24 22


[1] https://hub.docker.com/r/kartoza/geoserver/builds/

If you have time and inclination to merge it before the weekend, I will welcome your patches.

Regards

Tim

shane-axiom:
@timlinux Ok, I created PRs for all the building branches and also updated the master branch to use GS 2.8.2. Not sure about the `develop` and `2.8` branches, maybe they can be deleted?

PR #5, opened 2016-03-15
Incorporate master changes, set GS version to 2.5.2

shane-axiom:


timlinux:
Merged thanks @shane-axiom

PR #9, opened 2016-03-18
Add mkdir for GEOSERVER_DATA_DIR

daf:
This Dockerfile sets the ENV but doesn't create the directory, leading to geoserver outputting things like these in the logs:

```
ERROR [storage.DefaultStorageFinder] - Found System environment variable GEOSERVER_DATA_DIR set to /opt/geoserver/data_dir , but this path does not exist
...
WARN [geoserver.global] - Found System environment variable GEOSERVER_DATA_DIR set to /opt/geoserver/data_dir , but this path does not exist
...
INFO [geoserver.global] - Falling back to embedded data directory: /usr/local/tomcat/webapps/geoserver/data
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,8 @@ RUN apt-get -y update
 ENV GS_VERSION 2.8.2
 ENV GEOSERVER_DATA_DIR /opt/geoserver/data_dir
 
+RUN mkdir -p $GEOSERVER_DATA_DIR
+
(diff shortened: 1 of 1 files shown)


shane-axiom:
+1, I missed this. Not a show stopper since if a user isn't including a data directory as a volume or overlay GeoServer will revert back to using `./data` as the data dir, but it should be fixed.

timlinux:
Thanks for the fix @daf and @shane-axiom !

PR #11, opened 2016-05-31
fix typo in docker run commands

mathieu-pousse:


timlinux:
Hoooo yeah geosever sounds like a painful thing - thanks!

PR #14, opened 2016-09-28
Update to GeoServer 2.9.1, Java 8

shane-axiom:


timlinux:
Awesome, thanks very much @shane-axiom

PR #15, opened 2016-11-09
Add a logic to fetch local files of JAI and JAI ImageIO

seifer08ms:
Added a little logic that will fetch the JAI and JAI ImageIO tar files if they are not available locally in the resources folder.


timlinux:
Thanks very much @seifer08ms

PR #17, opened 2016-12-12
Fix for JAI/imageio download bug

peterataylor:
The wget commands need to put the JAI/imageio zips into the resources directory (if not available), otherwise subsequent mv command fails. Reproduce by doing a docker build without the local tars available.


timlinux:
Thanks very much @peterataylor - merged...

PR #19, opened 2017-05-31
Modified Dockerfile

bolshoydi:
while building were thrown an errors like this "mv: cannot stat ‘/tmp/gs_plugin/*.jar’: No such file or directory"


timlinux:
Thank you for this contribution!

Issue #21, opened 2017-12-10
Plugin Installation not working?


Boabob:
I changed the GS_Version within the Dockerfile to:
    ENV GS_VERSION 2.12.1

I downloaded one of the plugin files to:
    docker-geoserver/resources/plugins/geoserver-2.12.1-vectortiles-plugin.zip

Also downloaded:
    docker-geoserver/resources/geoserver.zip

Then built:
    docker build --build-arg TOMCAT_EXTRAS=false -t kartoza/geoserver .

And then started the GeoServer. It became version 2.12.1 but my pluging is not installed.

paulsenne:
Had this same issue. The block from 93-100 has the path to the plugins in double quotes and my system did not like this with a wildcard. Removing the double quotes works for me.

Instead of
```sh
# Install any plugin zip files in resources/plugins
RUN if ls "/tmp/resources/plugins/*.zip" > /dev/null 2>&1; then \
      for p in "/tmp/resources/plugins/*.zip"; do \
        unzip $p -d "/tmp/gs_plugin" \
        && mv "/tmp/gs_plugin/*.jar" $CATALINA_HOME/webapps/geoserver/WEB-INF/lib/ \
        && rm -rf /tmp/gs_plugin; \
      done; \
    fi
```

this works for me
```sh
# Install any plugin zip files in resources/plugins
RUN if ls /tmp/resources/plugins/*.zip > /dev/null 2>&1; then \
      for p in /tmp/resources/plugins/*.zip; do \
        unzip $p -d /tmp/gs_plugin \
        && mv /tmp/gs_plugin/*.jar $CATALINA_HOME/webapps/geoserver/WEB-INF/lib/ \
        && rm -rf /tmp/gs_plugin; \
      done; \
    fi
```

MattF-NSIDC:
https://github.com/kartoza/docker-geoserver/pull/25

timlinux:
Thanks for the fix @MattF-NSIDC

PR #25, opened 2018-01-30
Fix plugin install globbing bug

MattF-NSIDC:
When asterisks are quoted, they are treated as literal!


MattF-NSIDC:
Addresses https://github.com/kartoza/docker-geoserver/issues/21

timlinux:
Thanks for this @MattF-NSIDC !

PR #28, opened 2018-02-25
Added extra options for geoserver

NyakudyaA:
* Added a download script to install plugins in Geoserver. The script should be run before building the image. Some common plugins should be preinstalled in the image (ie control-flow, GDAL, WPS, Image pyramid)
* Added GDAL data directory and environment variables to ensure when a user installs the Gdal plugins Geoserver recognizes them.
* Moved the java environment settings to the docker file so that the image contains built-in optimizations.
* Modified build script so that when building the image locally you only specify the GeoServer version once in the download script and it changes all occurrences in other files. Example Dockerfile and download.sh.
* Added more instructions in the readme including how to use a persistent storage with sample data from GeoServer war file.
* Added a sample file for use with the control flow module.
* Install libjpeg-turbo- An extension which provides a significant performance enhancement for JPEG encoding in GeoServer WMS output
* Install extra fonts which are used by SLD with graphic markers

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ ENV GEOSERVER_OPTS "-Djava.awt.headless=true -server -Xms2G -Xmx4G -Xrs -XX:Perf
 #-XX:+UseConcMarkSweepGC use this rather than parallel GC?  
 ENV JAVA_OPTS "$JAVA_OPTS $GEOSERVER_OPTS"
 ENV GDAL_DATA /usr/local/gdal_data
-ENV LD_LIBRARY_PATH /usr/local/gdal_native_libs:/usr/local/apr/lib
+ENV LD_LIBRARY_PATH /usr/local/gdal_native_libs:/usr/local/apr/lib:/opt/libjpeg-turbo/lib64
 ENV GEOSERVER_LOG_LOCATION /opt/geoserver/data_dir/logs/geoserver.log
 
 RUN mkdir -p $GEOSERVER_DATA_DIR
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -106,7 +153,7 @@ RUN rm -f /tmp/resources/overlays/README.txt && \
     fi;
 
 # Optionally remove Tomcat manager, docs, and examples
-ARG TOMCAT_EXTRAS=true
+#ARG TOMCAT_EXTRAS=true #moved to docker-compose.yml
 RUN if [ "$TOMCAT_EXTRAS" = false ]; then \
```
gubuntu: because the previous line has been moved to docker-compose, this will always be false and therefore run - delete it? Or should it run?

gubuntu:
LGTM - jsut address my one comment then merge (and test after merging)

README.md
```diff
@@ -15,19 +15,34 @@ get our docker trusted build like this:
...
+Inspect downloads.sh to confirm which files you want, then run `.downloads.sh.`
+
+If you don't make changes it will download Oracle Java and various Oracle and Geoserver extensions that will be used during the Docker build. 
+
+### Setting Tomcat properties during build
+
+The  Tomcat properties such as maximum heap memory size are included in the Dockerfile. You need to change them 
+them before building the image in accordance to the resources available on your server:
+
+The default settings in the image are:
```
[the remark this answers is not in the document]
NyakudyaA: I will delete that

Dockerfile
NyakudyaA: It should always be true. The arguments are substituted when building the image. So in the docker-compose, it would be 
```
    args:
      - ORACLE_JDK=true
      - TOMCAT_EXTRAS=false
```

PR #29, opened 2018-04-17
Added options to incoperate community modules in the build

NyakudyaA:
* Allow extra environmental variables to fine tune GeoServer - ie ENABLE_JSONP that allows for cross-domain utilization of WFS

* Build community modules and add them to the docker images. ie Geogig and Mondo database.

* Add a file for projections which allows users to add custom projections

* Add logfiles which allow users to have errors logged correctly in a case when a user does not specify an environment variable for the log file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ docker pull kartoza/geoserver
 ### Pre-downloading files
 
 Inspect downloads.sh to confirm which files you want, then run `.downloads.sh.`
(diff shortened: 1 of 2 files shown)


timlinux:
LGTM thanks @NyakudyaA 

We need to make a new branch for 2.13 when this is merged and setup docker hub to build that branch too...

PR #34, opened 2018-06-11
Add functions for Geonode 2.12.x and above

NyakudyaA:
*  Dockerfile that incorporates changes from [geoserver-rancher](https://github.com/geosolutions-it/geoserver-docker/blob/2.12.x/Dockerfile)
* Add all geonode instructions for building GeoServer to use with rancher
* Added build argument to separate building Geoserver for Geonode or just plain Geonode.

NB: This PR is meant to be reviewed so that we can talk to the geosolutions guys about the possibility of having a single image or repository for GeoServer. We still need to test if Docker cloud allows auto builds with arguments so that we can deprecate other branches


NyakudyaA:
Will close this PR. Need to test it first on rancher Geonode before deploying

PR #45, opened 2018-10-18
Optimize Dockerfile for final image size

avvertix:
Hey, thanks for the clean GeoServer Docker image. 

I was using it for a project and I realized that after the download is about 1 GB. Therefore I started thinking on how to reduce the final image size. Here is my attempt. As result on my machine Docker reports 655MB for the final image size.

This is just my way to say thank you, feel free to decide if the contribution make sense for you. Have a great day!


NyakudyaA:
Thanks @avvertix I will review this and comment / merge. One of our pipeline dreams is to use alpine as a base image because Ubuntu base images are generally big.

Dockerfile
```diff
@@ -2,60 +2,68 @@
...
 ADD resources /tmp/resources
 ADD scripts /scripts
 RUN chmod +x /scripts/*.sh
-RUN /scripts/setup.sh
 ADD scripts/controlflow.properties $GEOSERVER_DATA_DIR
 ADD scripts/sqljdbc4-4.0.jar $CATALINA_HOME/webapps/geoserver/WEB-INF/lib/
 
-# Clean up APT when done.
-RUN echo "Yes, do as I say!" | apt-get remove --force-yes sed
-RUN echo "Yes, do as I say!" | apt-get remove --force-yes login
```
NyakudyaA: Why did you remove this? Any reason

Dockerfile
```diff
@@ -2,60 +2,68 @@
...
 ARG ORACLE_JDK=false
+
+## Would you like to keep default Tomcat webapps
 ARG TOMCAT_EXTRAS=true
+
+## Would you like to install community modules
 ARG COMMUNITY_MODULES=true
 
+## Maximum Memory that Java can allocate
+ARG MAXIMUM_MEMORY="4G"
```
NyakudyaA: Nice Argument

Dockerfile
avvertix: Those needs to be executed earlier to have a real benefit, otherwise the previous layers will still contain the executables. I was not sure where `sed` or `login` were used in the previous steps

avvertix:
Yep, it was running when I tried. I will rebase on `develop` and resolve the conflicts

Dockerfile
avvertix: Added documentation in https://github.com/kartoza/docker-geoserver/pull/45/commits/9e9bcb8ca18d53d43d8cf3d7efce498a03f923c2

Dockerfile
avvertix: I added back the `login` package removal in https://github.com/kartoza/docker-geoserver/pull/45/files#diff-3254677a7917c6c01f55212f86c57fbfR33, but not `sed` as it was commented in `develop`

NyakudyaA: Thanks `sed` is now being used to substitute the new GeoServer admin password on runtime

avvertix:
I think I resolved all the conflicts and correctly rebased everything on `develop`.

I tried to build and run the Docker image and it is working. Maybe the original problem was caused by git cloning the shell scripts with CRLF line endings instead of LF.

Here how I tested it

```shell
docker build --pull -t avvertix-kartoza-geoserver .
docker run --name "geoserver"  -p 8080:8080 -d -t avvertix-kartoza-geoserver
```

image

NyakudyaA:
> I think I resolved all the conflicts and correctly rebased everything on `develop`.
> 
> I tried to build and run the Docker image and it is working. Maybe the original problem was caused by git cloning the shell scripts with CRLF line endings instead of LF.
> 
> Here how I tested it
> 
> ```shell
> docker build --pull -t avvertix-kartoza-geoserver .
> docker run --name "geoserver"  -p 8080:8080 -d -t avvertix-kartoza-geoserver
> ```
> 
> image

Awesome. I will just build it and run the container and if successful I will then merge.

NyakudyaA:
> > I think I resolved all the conflicts and correctly rebased everything on `develop`.
> > I tried to build and run the Docker image and it is working. Maybe the original problem was caused by git cloning the shell scripts with CRLF line endings instead of LF.
> > Here how I tested it
> > ```shell
> > docker build --pull -t avvertix-kartoza-geoserver .
> > docker run --name "geoserver"  -p 8080:8080 -d -t avvertix-kartoza-geoserver
> > ```
> > 
> > 
> > image
> 
> Awesome. I will just build it and run the container and if successful I will then merge.

@avvertix can you please check if your java memory setting is having an effect on the container.
screenshot 2018-10-30 at 16 50 38

And looking at the logs it shows that Tomcat APR is not loaded 
screenshot 2018-10-30 at 16 57 00

avvertix:
> if your java memory setting is having an effect on the container.

My changes did not affect how memory was managed before. I only parametrized the existing configuration. The memory showed in the status report seems not related to the JVM, but rather to [a specific imaging library](https://docs.geoserver.org/stable/en/user/configuration/image_processing/index.html)

My knowledge of Java is limited so I don't know how to verify that while running the JVM use those settings

> And looking at the logs it shows that Tomcat APR is not loaded

I'm not sure what change I made had this consequence. Probably an environment variable?
Right now I cannot check deeply, I will try to figure it out tomorrow, but I welcome any suggestion on what to check

NyakudyaA:
> > if your java memory setting is having an effect on the container.
> 
> My changes did not affect how memory was managed before. I only parametrized the existing configuration. The memory showed in the status report seems not related to the JVM, but rather to [a specific imaging library](https://docs.geoserver.org/stable/en/user/configuration/image_processing/index.html)
> 
> My knowledge of Java is limited so I don't know how to verify that while running the JVM use those settings
> 
> > And looking at the logs it shows that Tomcat APR is not loaded
> 
> I'm not sure what change I made had this consequence. Probably an environment variable?
> Right now I cannot check deeply, I will try to figure it out tomorrow, but I welcome any suggestion on what to check

I will look and make changes if I find it out. Will also look at the java thing when I have time.

avvertix:
I was able to find the change that had impact on the Tomcat APR: it was caused by the env variable `LD_LIBRARY_PATH` being entirely overwritten, while Tomcat docker image set in it the value of the APR native location

This was my run, hope that it will run the same on your machine

```log
31-Oct-2018 15:23:52.562 INFO [main] org.apache.catalina.core.AprLifecycleListener.lifecycleEvent Loaded APR based Apache Tomcat Native library 1.2.17 using APR version 1.5.2.
31-Oct-2018 15:23:52.562 INFO [main] org.apache.catalina.core.AprLifecycleListener.lifecycleEvent APR capabilities: IPv6 [true], sendfile [true], accept filters [false], random [true].
31-Oct-201
```

avvertix:
About the options of the JVM seems that the catalina startup script do not get the GeoServer options, I found that the log include a line about startup variables being used, and indeed there is no sign of the settings I'm making

avvertix:
I made a change to configure the JVM arguments within the `entrypoint.sh` script. This uses all environment variables configured at build time and made them available to the sub-shell where the catalina.sh script is executed. Probably most of the environment variables set at build time can be moved to the startup script, so they can be changed from the Docker run command or via Docker Compose without requiring to rebuild the whole image

Issue #51, opened 2018-11-09
Tomcat version is out of date


beteez:
The tomcat version the image is built on is pretty old (tomcat 7/jre-8). build options should be included to set those at build time.

NyakudyaA:
Closed via #https://github.com/kartoza/docker-geoserver/pull/53/commits/43b5df1bccb2fd046da5c0a1f065bcb252d8f80a

PR #52, opened 2018-11-16
Update Tomcat to 8.5 (i.e. 8-jre8)

nk027:
Seems to work out of the box on our instance.

Addresses #51.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 #--------- Generic stuff all our Dockerfiles should start with so we get caching ------------
-FROM tomcat:8.0-jre8
+FROM tomcat:8-jre8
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,5 +1,5 @@
 #--------- Generic stuff all our Dockerfiles should start with so we get caching ------------
-FROM tomcat:8.0-jre8
+FROM tomcat:8-jre8
```
NyakudyaA: Can you try the following
```
ARG image_version=8.0-jre8
FROM tomcat:${image_version}
```
@nk027 This will be better since the only change would be the build arguments

NyakudyaA:
@nk027 Please check my comments

NyakudyaA:
I have merged the changes into develop.

PR #55, opened 2019-02-06
Fix setup.sh

shane-axiom:
Lots of the downloads in setup.sh are broken. This PR fixes them, and causes setup.sh to exit if any of the commands fail (otherwise it will happily keep running after failing to download something and lead to a broken image)


NyakudyaA:
Thanks, @shane-axiom I hadn't updated master in a while but the develop branch has all the latest work I have been working on including reducing the image size after the build from 1gb to 600mb

PR #57, opened 2019-02-20
Fix problems arising from hardcoded jasypt path

mrpgraae:
When re-building the image with a newer Geoserver version, the `update_passwords.sh` script might fail, because the new Geoserver version uses a newer version of jasypt.

This PR fixes that, by not hardcoding the jasypt jar to 1.8, but instead looking for any jasypt jar located in `$CLASSPATH`.


NyakudyaA:
LGTM

PR #104, opened 2019-10-09
GeoServer 2.16.0

EHJ-52n:
- introduces geoserver user

- remove tomcate default apps by default

- increase GeoServer Version to 2.16.0

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ FROM tomcat:$IMAGE_VERSION
 LABEL maintainer="Tim Sutton<tim@linfiniti.com>"
 
 ## The Geoserver version
-ARG GS_VERSION=2.15.2
+ARG GS_VERSION=2.16.0
 
 ## Would you like to use Oracle JDK
 ARG ORACLE_JDK=false
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,6 +57,10 @@ ADD scripts/controlflow.properties $GEOSERVER_DATA_DIR
 ADD scripts/sqljdbc4-4.0.jar $CATALINA_HOME/webapps/geoserver/WEB-INF/lib/
 
 RUN /scripts/setup.sh \
+    && groupadd -r geoserver && useradd --no-log-init -r -g geoserver geoserver \
+    && chown --verbose --recursive geoserver:geoserver /opt/geoserver \
+    && chown --verbose --recursive geoserver:geoserver /opt/footprints_dir \
+    && chown --verbose --recursive geoserver:geoserver /usr/local/tomcat \
     && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*  \
     && dpkg --remove --force-depends  unzip
 
@@ -65,5 +69,7 @@ ENV \
     INITIAL_MEMORY="2G" \
     ## Maximum Memory that Java can allocate
     MAXIMUM_MEMORY="4G"
+
+USER geoserver
 
 CMD ["/scripts/entrypoint.sh"]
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -6,13 +6,13 @@ FROM tomcat:$IMAGE_VERSION
...
 ## The Geoserver version
-ARG GS_VERSION=2.15.2
+ARG GS_VERSION=2.16.0
 
 ## Would you like to use Oracle JDK
 ARG ORACLE_JDK=false
 
 ## Would you like to keep default Tomcat webapps
-ARG TOMCAT_EXTRAS=true
+ARG TOMCAT_EXTRAS=false
```
NyakudyaA: @EHJ-52n What is the reason for flagging this off since it is an argument which a user can built with or without

Dockerfile
```diff
@@ -57,6 +57,10 @@ ADD scripts/controlflow.properties $GEOSERVER_DATA_DIR
 ADD scripts/sqljdbc4-4.0.jar $CATALINA_HOME/webapps/geoserver/WEB-INF/lib/
 
 RUN /scripts/setup.sh \
+    && groupadd -r geoserver && useradd --no-log-init -r -g geoserver geoserver \
+    && chown --verbose --recursive geoserver:geoserver /opt/geoserver \
+    && chown --verbose --recursive geoserver:geoserver /opt/footprints_dir \
+    && chown --verbose --recursive geoserver:geoserver /usr/local/tomcat \
```
NyakudyaA: @EHJ-52n What is the difference between adding a user with a UID like
```
#change Container from Root User to geoserver
RUN groupadd -r geoserver -g 10001 && \
    useradd -M -u 10000 -g geoserver geoserver
RUN chown -R geoserver:geoserver /usr/local/tomcat && \
    chmod -R 511 $CATALINA_HOME/webapps/geoserver/WEB-INF/lib/
```

Dockerfile
EHJ-52n: okay, let's step back one step before we resolve this.

What are the tomcat extras / default apps?

- manager?
- documentation?
- examples?

If yes, I would recommend to use a tomcat docker image without them by default to save space and do not provide anything that goes beyond having a servlet container with geoserver in it.

What do you think about this?

Dockerfile
EHJ-52n: In terms of docker, each `RUN`, `ADD`, `USER`, `EXPOSE`,... command provides an additional layer which I would like to avoid when creating docker images.

In terms of user ids, what is the reason to use such high ids? Do you want to prevent using already existing ids of the docker host system?

In terms of chmod, why is mode `511` required for `geoserver/WEB-INF/lib/`?

Dockerfile
NyakudyaA: All tomcat docker images have this by default and for us, we have the flag that allows users to built an image with/ without is very nice. We can flag it off by default but we have generally kept it for users who need to administer the servlet through this interface

EHJ-52n: Okay, I will adjust my PR accordingly.

Dockerfile
NyakudyaA: I really do not know much about running containers as none root user so that's why I was just asking. I read here https://medium.com/@mccode/understanding-how-uid-and-gid-work-in-docker-containers-c37a01d01cf
https://medium.com/better-programming/running-a-container-with-a-non-root-user-e35830d1f42a

NyakudyaA:
@EHJ-52n I have been testing this branch and I know running docker images as root is not recommended but I have been getting errors on startup due to file permissions. Is there a way to fix this otherwise I will revert to using root in this container until we can figure out the full workflow

data-dir
access

PR #128, opened 2020-01-24
Added option for EXISTING_DATA_DIR

### LINKED ISSUE #127: update_passwords.sh is run unconditionally with negative side effects
`update_passwords.sh` is run unconditionally in the entrypoint with negative side effects when supplying an already existing data_dir.

Not sure if i'm doing it wrong but if i add a new data_dir and edit the `GEOSERVER_DATA_DIR`  in my Dockerfile the `update_scripts.sh` script will run and mangle my `$GEOSERVER_DATA_DIR/security/usergroup/default/users.xml` to be however many rows i had before but all duplicates of the changed admin password resulting in a nasty XML duplicate key error and localhost:8080/geoserver will never get online and tomcat throws a 404. 

If i remove the `update_password.sh` script from the entrypoint everything is working as expected. Is it possible to either add a condition to this script (ie. not running it if an admin row already exist), or fix the `sed` command to not mangle the file if it already has multiple users in it. 

This is only an issue when trying to migrate an already existing data_dir with a security folder.

NicolaiMogensen:
Fixes #127 by not running `update_passwords.sh` if the environment variable EXISTING_DATA_DIR is set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,11 @@ create the file at `resources/overlays/usr/local/tomcat/bin/setenv.sh`.
 You can use this functionality to write a static GeoServer directory to
 `/opt/geoserver/data_dir`, include additional jar files, and more.
 
+If you have an already existing `data_dir` with a security setup from another Geoserver: set `EXISTING_DATA_DIR=true`.
+This will keep the passwords from getting changed by docker. 
+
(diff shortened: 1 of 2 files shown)


NyakudyaA:
LGTM

PR #135, opened 2020-02-25
Fix sql server plugin installation

grzewal:


NyakudyaA:
> _No description provided._

Thanks but can you take a look at develop branch, it downloads all extensions and you can activate it using env variables . If the plugin is not added there please add it to https://github.com/kartoza/docker-geoserver/blob/develop/stable_plugins.txt

grzewal:
You are right. For some reason when I built the container the plugin for sqlserver was not downloaded... I will need to try building it again. At least now I can see that it should be downloaded according to the setup.sh

Issue #148, opened 2020-04-01
GEOSERVER_ADMIN_USER set to non-default doesn't get admin permission


stewartgordonsitka:
Hi!

First off, thank you for all of the hard work you've done on this. I'm very grateful, and I've had a lot of fun with this project.

We were having some slight issues with your "update_passwords.sh" script. When setting the GEOSERVER_ADMIN_USER env variable to something other than admin, we noticed that while the user was set up correctly with our corresponding GEOSERVER_ADMIN_PASSWORD env variable, the user didn't actually have admin access. After some digging around, it looks like the [update_passwords.sh](https://github.com/kartoza/docker-geoserver/blob/master/scripts/update_passwords.sh) script edits the geoserver users.xml file, but neglects to edit the geoserver roles.xml file to set the new GEOSERVER_ADMIN_USER as an admin.

We were able to work around this in our instance, but I wanted to let you know and submit a pull request with a fix that matches how you do the password setting.

Thanks again!

NyakudyaA:
@stewartgordonsitka thanks a lot I will check it out and test

PR #149, opened 2020-04-01
Edit the roles.xml file to have the custom geoserver admin user set as an admin

stewartgordonsitka:
Minor changes to the update_passwords.sh script to handle setting a custom admin username as an admin in geoserver. Regarding issue: #148


NyakudyaA:
LGTM

Issue #165, opened 2020-06-23
How do I upgrade from Apache Tomcat/9.0.20 to Apache Tomcat/9.0.36?


davepenton:
I want to upgrade to the latest patched version of Apache Tomcat. 

It looks like jre builds have come to an end, so I am trying with jdk. When I build with IMAGE_VERSION=9-jdk11-openjdk-slim, I get a couple of errors: 

The first creating a symbolic link:
`Step 9/26 : RUN set -e     export DEBIAN_FRONTEND=noninteractive     dpkg-divert --local --rename --add /sbin/initctl     && ln -s /usr/lib/jvm/java-11-openjdk-amd64 /usr/lib/jvm/default-java     && (echo "Yes, do as I say!" | apt-get remove --force-yes login)     && apt-get clean     && rm -rf /var/lib/apt/lists/*
 ln: failed to create symbolic link '/usr/lib/jvm/default-java': No such file or directory`

The second issue is configuring the webapps manager:
`cp /build_data/context.xml /usr/local/tomcat/webapps/manager/META-INF`

For my purposes, I can assume that the JAVA_HOME could be set explicitly to a new location (/usr/local/openjdk-11/bin/java). In the later release, webapps has moved to webapps.dist. Any webapps that are required have to be explicitly copied across. I don't think I need the manager, so I can remove this line. This leaves me with a patched Dockerfile ([update-to-9-jdk11-openjdk-slim.patch.txt](https://github.com/kartoza/docker-geoserver/files/4817672/update-to-9-jdk11-openjdk-slim.patch.txt)). However, I'm not sure if this would work for others though. 

So my query is: Have I missed something? Is there a better way to upgrade to the latest patch?

NyakudyaA:
Yes, you can explicitly set the java home to a different location. Even though you will not need the tomcat manager It will be good to include it for anyone using the image. You can do your PR and we will review it accordingly

davepenton:
Ok. I have created a PR for your review.

evilcins:
@NyakudyaA the proposed solution appears to work. Any chance you'll move this forward in the foreseeable time?

One other issue that I've came across is that the `stable_plugins.txt` are not installed, because the corresponding download endpoints don't work.

NyakudyaA:
> @NyakudyaA the proposed solution appears to work. Any chance you'll move this forward in the foreseeable time?

Yes, Can you change your PR to point to develop and I will merge.
> 
[...]

PR #166, opened 2020-06-25
Add support for Tomcat/9.0.36 #165

davepenton:
This is to support 9.0.36 without breaking earlier versions. I have moved definition of JAVA_HOME to an argument of the docker container. If webapps/manager is in the dist folder, it is copied across.


davepenton:
New pull request: https://github.com/kartoza/docker-geoserver/pull/170

PR #176, opened 2020-08-10
Docker image for SPC geonode 

NyakudyaA:
* Use kartoza recipe for Geonode install.  Based on logic from https://github.com/GeoNode/geonode/blob/master/scripts/spcgeonode/geoserver/docker-entrypoint.sh


NyakudyaA:
@lucernae Can you take a look here. We need to have our own geoserver docker recipe that e can use for SPC GeoNode deployments.

NyakudyaA:
@lucernae Are you going to review here, otherwise I will just merge

PR #195, opened 2020-10-26
Changes required to make this image work on openshift

VBP8501:
Hello, we needed a geoserver image that could run on openshift. I couldn't get the plain image to run on openshift because you don't know the user that will be running the services inside the container. Images running on open shift need to follow these guidelines: https://docs.openshift.com/container-platform/3.3/creating_images/guidelines.html

This pull request covers what is necessary to get it to work (tested on our local openshift platform).

Thank you for this repo!

Thanks to @kervel for the help!

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -e
 
+/scripts/openshift_entrypoint.sh
+
 /scripts/start.sh
 
 CLUSTER_CONFIG_DIR="${GEOSERVER_DATA_DIR}/cluster/instance_$RANDOMSTRING"
--- a/scripts/entrypoint_openshift.sh
+++ b/scripts/entrypoint_openshift.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+export USER_NAME=geoserveruser
+export HOME=/home/geoserveruser
+if ! whoami &> /dev/null; then
+  if [ -w /etc/passwd ]; then
+    echo "${USER_NAME:-default}:x:$(id -u):0:${USER_NAME:-default} user:${HOME}:/sbin/nologin" >> /etc/passwd
+  fi
+fi
+cd $HOME
--- a/Dockerfile
+++ b/Dockerfile
@@ -124,16 +124,19 @@ ENV \
 
 EXPOSE  $HTTPS_PORT
 
-RUN groupadd -r geoserverusers -g 10001 && \
-    useradd -m -d /home/geoserveruser/ --gid 10001 -s /bin/bash -G geoserverusers geoserveruser
-RUN chown -R geoserveruser:geoserverusers ${CATALINA_HOME} ${FOOTPRINTS_DATA_DIR}  \
- ${GEOSERVER_DATA_DIR} /scripts ${LETSENCRYPT_CERT_DIR} ${FONTS_DIR} /tmp/ /home/geoserveruser/ /community_plugins/ \
- /plugins
+RUN chmod g=u /etc/passwd && mkdir -p /home/geoserveruser
+
+
+RUN chgrp -R 0 ${CATALINA_HOME} ${FOOTPRINTS_DATA_DIR} \
+    ${GEOSERVER_DATA_DIR} /scripts ${LETSENCRYPT_CERT_DIR} ${FONTS_DIR} /tmp/ /home/geoserveruser/ /community_plugins/ /plugins && \
+    chmod -R g=u ${CATALINA_HOME} ${FOOTPRINTS_DATA_DIR} \
+    ${GEOSERVER_DATA_DIR} /scripts ${LETSENCRYPT_CERT_DIR} ${FONTS_DIR} /tmp/ /home/geoserveruser/ /community_plugins/ /plugins
 
 RUN chmod o+rw ${LETSENCRYPT_CERT_DIR}
 
-USER geoserveruser
 VOLUME ["${GEOSERVER_DATA_DIR}", "${LETSENCRYPT_CERT_DIR}", "${FOOTPRINTS_DATA_DIR}", "${FONTS_DIR}", "${GEOWEBCACHE_CACHE_DIR}"]
 WORKDIR ${CATALINA_HOME}
+
+USER 1001
 
 CMD ["/bin/sh", "/scripts/entrypoint.sh"]


NyakudyaA:
Thanks @VBP8501 , I will give it a test run in the morning

NyakudyaA:
@VBP8501 Can you address the issues I have mentioned in the review before I can test again

scripts/entrypoint.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -e
 
+/scripts/openshift_entrypoint.sh
```
VBP8501, line 4: this should have been `/scripts/openshift_entrypoint.sh`

VBP8501:
> @VBP8501 Can you address the issues I have mentioned in the review before I can test again

Sorry I'm not sure where I can find your review, could you please point me to there?

I had made a small mistake yesterday (see my review). I'll make the required change

NyakudyaA:
@VBP8501 I have tested the PR and it partially works.  if you run the following

`docker run -it -p 8600:8080 -u $(id -u)  -v `pwd`/geoserver_data:/opt/geoserver/data_dir kartoza/geoserver:2.18.0` It doesn't work because it expect the data directory to already exists hence it generates the following error

```
Adding custom projection directory
cp: cannot create directory '/opt/geoserver/data_dir/user_projections': Permission denied
/scripts/start.sh: line 188: /opt/geoserver/data_dir/controlflow.properties: Permission denied
cp: cannot create directory '/opt/geoserver/data_dir/security': Permission denied
cp: cannot stat '/opt/geoserver/data_dir/security/usergroup/default/users.xml': No such file or directory
/scripts/update_passwords.sh: line 36: /opt/geoserver/data_dir/security/usergroup/default/users.xml: No such file or directory
cat: /opt/geoserver/data_dir/security/usergroup/default/users.xml.orig: No such file or directory
cp: cannot stat '/opt/geoserver/data_dir/security/role/default/roles.xml': No such file or directory
/scripts/update_passwords.sh: line 41: /opt/geoserver/data_dir/security/role/default/roles.xml: No such file or directory
cat: /opt/geoserver/data_dir/security/role/default/roles.xml.orig: No such file or directory
touch: cannot touch '/opt/geoserver/data_dir/.updatepassword.lock': Permission denied
```
But if I create the geoserver_data directory before running the start command it actually works.
All files that are generated within the container have root access as well. ie 
gwc

What is the difference between the root access visible in the container. Is it different from the normal root users ?

kervel:
Hello, the "root" group is because openshift runs containers with a random userID (so there is no way you can assign the files to the right user, you simply don't know in advance which userID will be used by openshift). The only certainty you have is that the *group* of the user will be root. This is documented in the guidelines doc mentioned in the PR.

wrt the creation of directories: we didn't make any modification to the start.sh script, so are you sure there is a change wrt current behaviour ? i can do some more testing tomorrow. the modifications also don't do any user switching, permission denied seems strange. i will try to figure out what is going on tomorrow.

lucernae:
Hi, I was requested by @NyakudyaA to review this. Will take a look.

lucernae:
> Hello, the "root" group is because openshift runs containers with a random userID (so there is no way you can assign the files to the right user, you simply don't know in advance which userID will be used by openshift). The only certainty you have is that the group of the user will be root. This is documented in the guidelines doc mentioned in the PR.

We are not familiar with openshift, so we don't really know what the expected best practises are. However most of our docker image user use this image to transition from legacy monolithic setup into a microservice based deployment. In most cases, the Geoserver data directory already exists in the host file system, and bind mounted into the container, overlaying the dirpath inside the docker image. So it definitely will have different uid/gid. From what you describe, then there's no way to reconcile this case since we actually don't know what uid/gid the existing host folder will have.

In that case, can you suggest possible approach? Can we create an entrypoint scripts that automatically assign the folder permission to the correct groupt **at runtime**? Will that work? My hunch says it's not possible since the script inside already assigned with random UID which doesn't have power to change folder permission.

> wrt the creation of directories: we didn't make any modification to the start.sh script, so are you sure there is a change wrt current behaviour ? i can do some more testing tomorrow. the modifications also don't do any user switching, permission denied seems strange. i will try to figure out what is going on tomorrow.

Sorry, to be clear, what @NyakudyaA did was testing the new image with **existing host folder** made by the previous image. This is just to test if the new image will break/not run if our user (people who uses the image) upgrade the image in their own environment/deployment. To summarize, it's not your script at fault. It's just that maybe the script doesn't expect this kind of usage, so we have to find a way to handle it gracefully. Which is, to correct the folder permission at runtime so it uses proper group.

scripts/entrypoint.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -e
 
+/scripts/entrypoint_openshift.sh
```
lucernae, line 4: CMIIW, if we use this approach then this setup is not specific to openshift only. I think a more general name can do. Such as :`passwd_setup.sh`

scripts/entrypoint_openshift.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+export USER_NAME=geoserveruser
+export HOME=/home/geoserveruser
+if ! whoami &> /dev/null; then
+  if [ -w /etc/passwd ]; then
+    echo "${USER_NAME:-default}:x:$(id -u):0:${USER_NAME:-default} user:${HOME}:/sbin/nologin" >> /etc/passwd
+  fi
+fi
+cd $HOME
```
lucernae, line 9: The `cd` at the end was not used, I think. So we can omit it.

scripts/entrypoint_openshift.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+export USER_NAME=geoserveruser
+export HOME=/home/geoserveruser
+if ! whoami &> /dev/null; then
+  if [ -w /etc/passwd ]; then
```
lucernae, line 5: Should also test if the passwd contains the `$USER_NAME` entry.
In some cases, our user restart the container without deleting it. In those case, the passwd file will have duplicate entries.

lucernae (commented):
This is my review on the code itself.

scripts/entrypoint.sh
kervel: valid point, thanks, will change.

scripts/entrypoint_openshift.sh
kervel: ahh yes, that's indeed something we definately need to change. will check

scripts/entrypoint_openshift.sh
kervel: > > Hello, the "root" group is because openshift runs containers with a random userID (so there is no way you can assign the files to the right user, you simply don't know in advance which userID will be used by openshift). The only certainty you have is that the group of the user will be root. This is documented in the guidelines doc mentioned in the PR.
> 
> We are not familiar with openshift, so we don't really know what the expected best practises are. However most of our docker image user use this image to transition from legacy monolithic setup into a microservice based deployment. In most cases, the Geoserver data directory already exists in the host file system, and bind mounted into the container, overlaying the dirpath inside the docker image. So it will definitely have different uid/gid. From what you describe, then there's no way to reconcile this case since we actually don't know what uid/gid the existing host folder will have.

normally this is solved using group permissions at openshift. i think it also sets the umask accordingly (but i didn't find a reference in the openshift docs).
that made me think, i should add the 10001 user back so that by default it runs as 10001 and is compatible with the current docker images in non-openshift context.


> In that case, can you suggest possible approach? Can we create an entrypoint scripts that automatically assign the folder permission to the correct groupt **at runtime**? Will that work? My hunch says it's not possible since the script inside already assigned with random UID which doesn't have power to change folder permission. But I maybe wrong.

that would not work as if the openshift user has no write access, nothing can be done at runtime to fix it ...

> > wrt the creation of directories: we didn't make any modification to the start.sh script, so are you sure there is a change wrt current behaviour ? i can do some more testing tomorrow. the modifications also don't do any user switching, permission denied seems strange. i will try to figure out what is going on tomorrow.
> 
> Sorry, to be clear, what @NyakudyaA did was testing the new image with **existing host folder** made by the previous image. This is just to test if the new image will break/not run if our user (people who uses the image) upgrade the image in their own environment/deployment. To summarize, it's not your script at fault. It's just that maybe the script doesn't expect this kind of usage, so we have to find a way to handle it gracefully. Which is, to correct the folder permission at runtime so it uses proper group.
> 
> Can you comment on my question @kervel @VBP8501 ?

scripts/entrypoint_openshift.sh
kervel: actually the test for the $USER_NAME entry is already there, when $USER_NAME is not in /etc/passwd the "whoami" command fails and that enters the if.

lucernae: Oh yes, you are right. It's not that obvious to me.

bstrdsmkr:
I think this may no longer be valid, I was just able to deploy the helm chart v0.3.3 into Openshift without any issues

NyakudyaA:
Awesome, we can close the PR and reopen when needed

MathiasVDA:
> I think this may no longer be valid, I was just able to deploy the helm chart v0.3.3 into Openshift without any issues

that's right, we've also got it running on our openshift cluster

bstrdsmkr:
@NyakudyaA I apologize, I was wrong about this, these changes ARE still needed. The image will launch on OpenShift, but none of the dynamic functionality (plugins, config editing, etc) will work without these changes. In addition, I had to add the /settings directory to the list of chown'd directories

PR #213, opened 2021-01-27
Allow environment variables to be set from Docker secret files (similar to kartoza/docker-postgis)

MichaelFerranteSitka:
We have found that sometimes some of the environment variables set to get GeoServer running have sensitive information in them like credentials or passwords. Having a way to set those environment variables from Docker secret files can be helpful. This code follows a pattern we found in another kartoza GitHub repository kartoza/docker-postgis.


NyakudyaA:
Thanks @MichaelFerranteSitka see my comments

PR #225, opened 2021-03-11
Allow specifying arbitrary UID/GID for geoserveruser

ngilles:
Allow specifying arbitrary UID/GID for geoserveruser as some existing geoserver data dirs might have existing uid/gid ownership that can't be changed.


NyakudyaA:
Thanks @ngilles Can you please populate the Readme with how this will work.

ngilles:
Hi @NyakudyaA, I have to say, that was a bit rough of pull request! I've added the args to the .env and a short description in the readme in the build args section.

PR #241, opened 2021-04-22
.env: fix uid/gid variable names

dunn:


dunn:
oops, retargeting it from the github ui was not a good idea; i'll close this and open a new pr

PR #242, opened 2021-04-23
Expand SSL params to enable override of maxHttpHeaderSize

andyphenix:
Addresses issues we ran into with long CQL filters


NyakudyaA:
LGTM

PR #256, opened 2021-06-05
Random password

### LINKED ISSUE #237: cannot load custom fonts
I am trying to load custom fonts to use in my sld labels but it seems that none of my loaded fonts are being installed.

I am using the ` kartoza/geoserver:2.18.0` image. And here is how I am trying to add the fonts:

I copied the `.ttf` files into a `fonts` directory and I mount this volume in docker-compose. Here is the geoserver service as defined in my docker-compose file:

```yml
  geoserver:
    image: kartoza/geoserver:2.18.0
    ports:
      - 8600:8080
    env_file: .env
    volumes:
      - ./fonts:/opt/fonts
      - geoserver:/opt/geoserver/data_dir
    restart: always
    depends_on:
      - postgres
```

The contents of my `fonts` directory are:

```bash
ls fonts
NotoSans-Bold.ttf       NotoSans-BoldItalic.ttf NotoSans-Italic.ttf     NotoSans-Regular.ttf
```

And in the docker container (after connecting via `docker-compose exec geoserver bash`) the fonts are copied to the `$FONTS_DIR`:

```bash
geoserveruser@e052ba9c7749:/usr/local/tomcat$ ls ${FONTS_DIR}/*.ttf
/opt/fonts/NotoSans-Bold.ttf  /opt/fonts/NotoSans-BoldItalic.ttf  /opt/fonts/NotoSans-Italic.ttf  /opt/fonts/NotoSans-Regular.ttf
```

However, dont see these in the `truetype` directory:

```bash
geoserveruser@e052ba9c7749:/usr/local/tomcat$ ls /usr/share/fonts/truetype/
aenigma  dejavu  fonts-georgewilliams  msttcorefonts  sjfonts  ttf-bitstream-vera
```

When I observe the docker-compose logs, I notice that there are permissions issues:

```bash
 docker-compose logs geoserver     
Attaching to internal_geoserver_1
geoserver_1     | cp: cannot create regular file '/usr/share/fonts/truetype/NotoSans-Bold.ttf': Permission denied
geoserver_1     | cp: cannot create regular file '/usr/share/fonts/truetype/NotoSans-BoldItalic.ttf': Permission denied
geoserver_1     | cp: cannot create regular file '/usr/share/fonts/truetype/NotoSans-Italic.ttf': Permission denied
geoserver_1     | cp: cannot create regular file '/usr/share/fonts/truetype/NotoSans-Regular.ttf': Permission denied
geoserver_1     | Adding custom projection directory
geoserver_1     | Installing default data directory
...
```

Is this somehow related to Perhaps this is related to https://github.com/kartoza/docker-geoserver/issues/198 ? The files are not owned by root.

Am I missing a crucial step to use custom fonts? 

Thank you!

### LINKED ISSUE #244: Please add configuration for InitiatingHeapOccupancyPercent
I have a system that was recently launched that is running the latest docker w/Geoserver 2.18.2.  I also have a couple systems that are running an older version (Geoserver 2.16.2 built a year or so ago).  All of the systems are running in Kubernetes on a dedicated host with 8GB of ram and Geoserver is configured with MAXIMUM_MEMORY of 4GB.  We have been experiencing out of memory errors on the new system after it has been running for 2-3 days while the older systems have been running fine for many months with no issues at all.

While under load, the memory usage would steadily increase and never be reclaimed - even when usage dropped at the end of the day.  Eventually, the JVM would get an OOM error which crashes the container and causes it to restart.  It is not a Kubernetes pod eviction.

The changes to the garbage collection to switch from UseParallelGC to UseG1GC seems to be the difference.  And the only way I could get things working properly was to change the setting for InitiatingHeapOccupancyPercent back to Java's default of 45%.  

I'm not very experienced with Java (so correct me if I have this wrong) but from what I have learned in the last couple days, it seems that G1GC requires multiple full GC's to actually reclaim any memory.  So I think what's happening is when it hits 70% (which is what you set it to), it marks the memory as not allocated in the JVM heap but doesn't actually free it.  So it continues to allocate more memory until it...can't...because another GC is not triggered in time due to the heap being reduced by the previous pass.

I'm sure you had a reason for changing it to 70%, but could you please add a configuration for it so we can tweak it for our environments along with the initial/max settings?

Here is a picture of the cpu & memory usage of the VM from this week.  Over the weekend, I had forced Geoserver to free memory (from the status page) so it starts out low on Mon.  The 1st big drop in memory usage (and spike in CPU) is where the container crashed and restarted.  The 2nd is where I updated it (last night) to set InitiatingHeapOccupancyPercent to 45.  You can see it has been allocating and releasing memory normally today (even while under load) where it never released any significant amount of memory prior to that change.

image

### LINKED ISSUE #247: Expanding support of specifying passwords with _FILE?
Regarding the appending of `_FILE` to certain settings to make use of docker secrets, the readme file states that:

>Currently, GEOSERVER_ADMIN_USER and GEOSERVER_ADMIN_PASSWORD are supported.

Will the other password settings in the `.env` file be able to be configured using docker secrets in the near future? I don't feel comfortable specifying these settings in `.env` since I would like to version control my `.env` file.

The passwords I'm referring to are:

- `JKS_KEY_PASSWORD`
- `JKS_STORE_PASSWORD`
- `PKCS12_PASSWORD`

### LINKED ISSUE #249: Configuration when running behind nginx proxy?
I am trying to run geoserver on a server (for example, `server.backend.co.za`) that is behind a firewall, and where a separate nginx proxy server (for example, running as `server.frontend.co.za`) handles incoming requests.

The nginx maps calls from `/prefix/geoserver` to the http://server.backend.co.za:8124/geoserver/

This works fine for the title page; however all the data links are still shown as:

http://server.backend.co.za:8124/geoserver/tiger/wms?service=WMS&version=1.1.0&request=GetMap&layers=tiger... 

I have tried adding the https proxy details in the `docker-compose.yml` file:
```
        - HTTPS_PROXY_NAME=server.frontend.co.za
        - HTTPS_PROXY_PORT=80
```
but this had no effect.

What should be changed or set in docker-compose to enable the correct server prefix for these links?

NyakudyaA:
Fixes #237 
Fixes #247 
Fixes #249 
Fixes #244


andyphenix:
Running into an issue with these changes...the setting of a random password should be done after the check for a _FILE value (& only if that value not supplied):

https://github.com/kartoza/docker-geoserver/blob/fa436963769aa317581e8f8ba3c9b573cbc0a387/scripts/update_passwords.sh#L32

NyakudyaA:
> Running into an issue with these changes...the setting of a random password should be done after the check for a _FILE value (& only if that value not supplied):
> 
> https://github.com/kartoza/docker-geoserver/blob/fa436963769aa317581e8f8ba3c9b573cbc0a387/scripts/update_passwords.sh#L32

@andyphenix Can you do a PR for this otherwise I can do it when I am free

andyphenix:
Done - see https://github.com/kartoza/docker-geoserver/pull/262

PR #260, opened 2021-06-12
Fix some small bugs in mounting files

NyakudyaA:


andyphenix:
Running into an issue with these changes... 'advertise' --> command not found 

https://github.com/kartoza/docker-geoserver/blob/master/scripts/entrypoint.sh#L10

NyakudyaA:
@andyphenix That should not be there, will do a PR shortly to remove that line

PR #303, opened 2021-09-21
Add HTTP_SCHEME environment variable 

mandalorian-one:
Add HTTP_SCHEME environment variable to configure "scheme" attribute for "Connector" section in tomcat server.xml config file.
This is necessary in order to server constructs URLs correctly for reverse proxies for the case when reverse proxy serves HTTPS connection but GeoServer's backend serves HTTP connection. Without scheme's attribute set to "https" geoserver creates incorrect URLs with http scheme instead https


NyakudyaA:
Ping @mandalorian-one

mandalorian-one:
hi
is this PR acceptable? need some clarification?

NyakudyaA:
> hi
> is this PR acceptable? need some clarification?

@mandalorian-one  

Don't we need to set

```
if [ -z ${HTTP_SCHEME} ]; then
    HTTP_SCHEME=http
fi
```
in the `env-data.sh`

PR #312, opened 2021-10-15
282 Remove Volume definition in Dockerfile and apply persistent behaviour to docker-compose

### LINKED ISSUE #282: Remove `VOLUME` definition in Dockerfile
Since the PR of https://github.com/kartoza/docker-geoserver/pull/193/commits `VOLUME` have been defined in the Dockerfile. I would advocate removing this definition for the following reason:

Using the `VOLUME` definition, in my opinion, only makes sense when you are using Geoserver as a service through the UI. Where you want the run-time changes to be persistent using docker volumes. In cases where you would use Geoserver UI-less and want a stateless image that uses the `COPY` definition to embed the configs, it does _not_ make sense to use volumes. This last scenario would currently not work as the volumes would simply overwrite any embedded state.

I instead suggest to solely rely on the docker-compose `volume:` definitions to define persistent volumes for the docker run. This way the user of this Geoserver image can decide if they want this volume behaviour or not.

Please let me know what you think about this and if I missed something.

----

Some more discussion about this in the following issue: https://github.com/moby/moby/issues/3465

Nkmol:
This Pull Request closes #282. It makes sure that no `VOLUME` is defined inside of the Dockerfile so that the user of this image has full control over how they want to define persistency in their Geoserver stack. 

This makes it possible for someone to allow persistency at built vs persistency at runtime (see the ticket for a better explanation). 

----

I changed the docker-compose to have the same behaviour as the `VOLUME` did. As this docker-compose is just a suggestion for someone using this image, I do not believe this is that important.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,7 +66,7 @@ EXPOSE  $HTTPS_PORT
 
 USER ${GEOSERVER_UID}
 RUN echo 'figlet -t "Kartoza Docker GeoServer"' >> ~/.bashrc
-VOLUME ["${GEOSERVER_DATA_DIR}", "${CERT_DIR}", "${FOOTPRINTS_DATA_DIR}", "${FONTS_DIR}"]
+
 WORKDIR ${GEOSERVER_HOME}
 
 CMD ["/bin/bash", "/scripts/entrypoint.sh"]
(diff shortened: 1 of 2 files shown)


NyakudyaA:
ping @Nkmol

Nkmol:
@NyakudyaA I must be blind, but I do not see any feedback/comments.

afbeelding

NyakudyaA:
Here it is image

Nkmol:
@NyakudyaA Thanks for that, no idea why I can't see this comment. 

I completely agree with this comment as I also was not sure if we should have this behaviour. I removed the extra volume declarations for now.

PR #354, opened 2021-12-16
Update Geoserver to 2.20.2 for CVE-2021-44228 and CVE-2019-17571

adonm:
Update to 2.20.2 for http://geoserver.org/announcements/2021/12/13/logj4-rce-statement.html patches regarding  CVE-2021-44228 and CVE-2019-17571

### DIFF
--- a/.env
+++ b/.env
@@ -1,11 +1,11 @@
 COMPOSE_PROJECT_NAME=kartozageoserver
 
 IMAGE_VERSION=9.0-jdk11-openjdk-slim-buster
-GS_VERSION=2.20.1
+GS_VERSION=2.20.2
 GEOSERVER_PORT=8600
 # Build Arguments
 JAVA_HOME=/usr/local/openjdk-11
-WAR_URL=http://downloads.sourceforge.net/project/geoserver/GeoServer/2.20.1/geoserver-2.20.1-war.zip
+WAR_URL=http://downloads.sourceforge.net/project/geoserver/GeoServer/2.20.2/geoserver-2.20.2-war.zip
 DOWNLOAD_ALL_STABLE_EXTENSIONS=1
 DOWNLOAD_ALL_COMMUNITY_EXTENSIONS=1
 GEOSERVERUSER_UID=1000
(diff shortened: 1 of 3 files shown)


NyakudyaA (changes requested):
Please also change https://github.com/kartoza/docker-geoserver/blob/master/.env#L8 and https://github.com/kartoza/docker-geoserver/blob/master/.env#L4
Remove the logic https://github.com/kartoza/docker-geoserver/blob/master/scripts/setup.sh#L171
 This was used to update 2.20.1 and latest to patch that vulnerability

lucernae:
@NyakudyaA I think it is better to keep your patching logic in  https://github.com/kartoza/docker-geoserver/blob/master/scripts/setup.sh#L171 . We could patch older version of GeoServer image at runtime using that script.

NyakudyaA:
> @NyakudyaA I think it is better to keep your patching logic in https://github.com/kartoza/docker-geoserver/blob/master/scripts/setup.sh#L171 . We could patch older version of GeoServer image at runtime using that script.

@lucernae maybe we move the logic to start.sh so that it fixes older images on run time instead of setup stage

adonm:
Note I have tweaked as suggested (except for the patching logic), however until 2.20.2 is published to sourceforge the Dockerfile won't build - once they are published this request could be sensibly reviewed for a merge

NyakudyaA (approved):
Thanks, @adonm Let's wait till the new version is available before we merge this.

jt-helsinki:
Can you please check that all extensions are working before approving? MBStyles for example does not work. Version problem.

NyakudyaA:
> Can you please check that all extensions are working before approving? MBStyles for example does not work. Version problem.

@jt-helsinki can you open a new issue for this. We would rather have this addressed in a separate PR. Also describe what you mean by saying it doesn't work. Does it not install or the configuration is broken ?

NyakudyaA:
@adonm Can you please pull the latest changes from Master and fix the conflicts. Alternatively, can you also update https://github.com/kartoza/docker-geoserver/blob/master/scripts/setup.sh#L170 to have the following `if [[  -f ${GEOSERVER_INSTALL_DIR}/webapps/geoserver/WEB-INF/lib/log4j-1.2.17.jar ]] && [[ "${GS_VERSION}" -lt 2.20.2  ]] ; then`

NyakudyaA:
merged a PR to fix this and other missing commits from develop

PR #355, opened 2021-12-16
fix build url for older versions

NyakudyaA:
* Add build arg `STABLE_PLUGIN_BASE_URL=https://liquidtelecom.dl.sourceforge.net` because older archived versions fetched plugins from https://master.dl.sourceforge.net

We need to build the following images
* 2.16.5 , 2.17.5, 2.18.5 with the build ARG  `STABLE_PLUGIN_BASE_URL=https://master.dl.sourceforge.net`
* 2.19.4,2.20.2 should build fine with the current scripts and args

### DIFF
--- a/.env
+++ b/.env
@@ -6,6 +6,7 @@ GEOSERVER_PORT=8600
 # Build Arguments
 JAVA_HOME=/usr/local/openjdk-11
 WAR_URL=http://downloads.sourceforge.net/project/geoserver/GeoServer/2.20.1/geoserver-2.20.1-war.zip
+STABLE_PLUGIN_BASE_URL=https://liquidtelecom.dl.sourceforge.net
 DOWNLOAD_ALL_STABLE_EXTENSIONS=1
 DOWNLOAD_ALL_COMMUNITY_EXTENSIONS=1
 GEOSERVERUSER_UID=1000
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ FROM tomcat:$IMAGE_VERSION
 LABEL maintainer="Tim Sutton<tim@linfiniti.com>"
 ARG GS_VERSION=2.20.1
 ARG WAR_URL=https://downloads.sourceforge.net/project/geoserver/GeoServer/${GS_VERSION}/geoserver-${GS_VERSION}-war.zip
+ARG STABLE_PLUGIN_BASE_URL=https://liquidtelecom.dl.sourceforge.net
 ARG DOWNLOAD_ALL_STABLE_EXTENSIONS=1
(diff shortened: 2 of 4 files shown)


lucernae:
@NyakudyaA what about the original URL: https://tenet.dl.sourceforge.net is it not used anymore?

NyakudyaA:
> @NyakudyaA what about the original URL: https://tenet.dl.sourceforge.net is it not used anymore?

They do change these urls a lot.

lucernae:
@NyakudyaA I think it's best if I put the config in GH Action for this STABLE_PLUGIN_BASE_URL to be empty so it directly
use what is described in the Dockerfile

NyakudyaA:
> @NyakudyaA I think it's best if I put the config in GH Action for this STABLE_PLUGIN_BASE_URL to be empty so it directly
> 
> use what is described in the Dockerfile

The older images will not build using the default url in Dockerfile. I think the download of plugins will need to be adjusted to have an exit rule when the download fails so that the build completely fails

lucernae:
> The older images will not build using the default url in Dockerfile. I think the download of plugins will need to be adjusted to have an exit rule when the download fails so that the build completely fails

Roger that. I think I misunderstood the earlier post.
Can I summarize it as this:
- Same Dockerfile in a branch should be able to be used for building all stated images above
- Build args passed for STABLE_PLUGIN_BASE_URL depends on the above rule, since the env inside the Dockerfile is going to track for the most recent one, that means STABLE_PLUGIN_BASE_URL needs to be passed before the build

My problem at the moment is just that the build matrix rule in GitHub action becomes a little bit complicated and isn't too flexible. I will try to patch things up for now.

Should we implement the fail early logic in the Dockerfile in this repo or you are planning to merge it first?

NyakudyaA:
> > The older images will not build using the default url in Dockerfile. I think the download of plugins will need to be adjusted to have an exit rule when the download fails so that the build completely fails
> 
> 
> 
> Roger that. I think I misunderstood the earlier post.
> 
> Can I summarize it as this:
> 
> - Same Dockerfile in a branch should be able to be used for building all stated images above
> 
> - Build args passed for STABLE_PLUGIN_BASE_URL depends on the above rule, since the env inside the Dockerfile is going to track for the most recent one, that means STABLE_PLUGIN_BASE_URL needs to be passed before the build

You have explained it correctly. 
> 
> 
> 
> My problem at the moment is just that the build matrix rule in GitHub action becomes a little bit complicated and isn't too flexible. I will try to patch things up for now.

I understand what you mean but unless we figure out at what point does GeoServer change it's URL for archived versions then we can't really use the same URL for previous versions.
> 
> 
> 
> Should we implement the fail early logic in the Dockerfile in this repo or you are planning to merge it first?

I think you can merge first. This can be fixed in follow up PR.

PR #359, opened 2021-12-31
fix sed substitution for multiple users in the users and role configs

NyakudyaA:
* Default behaviour only works when an admin user is singular, when multiple users are added it was replacing all credentials


NyakudyaA:
@lucernae Can you take a look at this logic

PR #362, opened 2022-01-10
Adding variables for language, region, and country

curtis18:
Adding variables for language, region, and country and set geotools.localDateTimeHandling to true


NyakudyaA:
LGTM

PR #369, opened 2022-01-25
re-add wps-plugin in stable plugins

eblondel:
It seems the wps-plugin has been removed with https://github.com/kartoza/docker-geoserver/pull/358 and replaced by wps-download-plugin, although it's a different plugin. See it listed in the Services extensions at http://geoserver.org/release/stable/

### DIFF
--- a/build_data/stable_plugins.txt
+++ b/build_data/stable_plugins.txt
@@ -32,6 +32,7 @@ teradata-plugin
 wcs2_0-eo-plugin
 web-resource-plugin
 wmts-multi-dimensional-plugin
+wps-plugin
 wps-cluster-hazelcast-plugin
 wps-download-plugin
 wps-jdbc-plugin


NyakudyaA (commented):
The WPS plugin is installed by default in https://github.com/kartoza/docker-geoserver/blob/master/scripts/setup.sh#L64

eblondel:
I'm running the default config of the image and with the recent change, the WPS-plugin was not anymore installed (was in the past), i had to add it manually as environment param.

eblondel:
See how i had to add the wps-plugin 
 ​docker run --link postgis:postgis -d -p 8080:8080 -e SAMPLE_DATA=true -e STABLE_EXTENSIONS=wps-plugin kartoza/geoserver:latest

Without the explicit mention of wps-plugin, it is not installed anymore in the default image

NyakudyaA:
> See how i had to add the wps-plugin ​docker run --link postgis:postgis -d -p 8080:8080 -e SAMPLE_DATA=true -e STABLE_EXTENSIONS=wps-plugin kartoza/geoserver:latest
> 
> Without the explicit mention of wps-plugin, it is not installed anymore in the default image

I will update the latest image when https://github.com/kartoza/docker-geoserver/pull/354 is merged and this should sort it out. Use the image 2.20.1

NyakudyaA:
Please use the latest image 2.20.2 or latest as it has this fixed

PR #394, opened 2022-04-14
Upgrade GeoServer from version 2.20.3 to 2.20.4 (closes #393)

### LINKED ISSUE #393: Upgrade Geoserver to 2.20.4 to close Spring RCE
Geoserver 2.20.3 uses the vulnerable (https://tanzu.vmware.com/security/cve-2022-22965) Spring version 5.1.20.RELEASE. Can you please release a new version with the upgraded version 2.20.4?
If it is as simple as changing 2.20.3 to 2.20.4 in all places in your code I can prepare a PR, but I fear that there might be more to it.

stranljip:
This PR closes #393 and fixes the upgrade script so that it can be used with MacOS.

After the PR nothing in the behavior of Geoserver should be changed. Since there is a minor version update of Spring included in the patch version update from Geoserver 2.20.3 to 2.20.4 one should be attentive for bug reports in the Geoserver project concerning this change.

### DIFF
--- a/.env
+++ b/.env
@@ -1,11 +1,11 @@
 COMPOSE_PROJECT_NAME=kartozageoserver
 
 IMAGE_VERSION=9.0-jdk11-openjdk-slim-buster
-GS_VERSION=2.20.3
+GS_VERSION=2.20.4
 GEOSERVER_PORT=8600
 # Build Arguments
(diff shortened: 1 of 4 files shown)


NyakudyaA:
LGTM

PR #395, opened 2022-04-25
correction on sed command

iamtekson:


NyakudyaA:
I think this was reverted in https://github.com/kartoza/docker-geoserver/pull/394/files. I will test this on Mac as well as linux

PR #411, opened 2022-06-22
Parametrize shapefile datetime format

mawoc:
This change allows to disable datetime format in geoserver shp handling (wfs shape-zip output).


NyakudyaA:
LGTM

Issue #425, opened 2022-07-11
Docker build error


gutleber:
building last version failed on libgdal-java
exit code: 100
The following packages have unmet dependencies:
  libgdal26 : Depends: libarmadillo9 (>= 1:9.200.7+dfsg) but it is not installable
              Depends: libcfitsio7 (>= 3.450~) but it is not installable
              Depends: libdap25 but it is not installable
              Depends: libjson-c3 (>= 0.11) but it is not installable
              Depends: libnetcdf13 (>= 1:4.0.0) but it is not installable
              Depends: libpoppler82 (>= 0.71.0) but it is not installable
              Depends: libqhull7 but it is not installable
Unable to correct problems, you have held broken packages.

How can I resolve this issue?

NyakudyaA:
Are you building this manually? Are you using master or develop branch

gutleber:
"docker-compose -f docker-compose-build.yml up -d --build" also end with same result

NyakudyaA:
I have a feeling the PPA we use might have also been updated to remove libgdal. If that's the case we might have to try another PPA or alternatively downgrade to use a lower version of gdal. This issue of dropping the java bindings is discussed on various platforms i.e https://trac.osgeo.org/osgeolive/ticket/2288

NyakudyaA:
I tried building this and it worked. Are you not able to pull the image from the docker hub repository? Otherwise you will need to checkout commit https://github.com/kartoza/docker-geoserver/commit/162c562d0cec6eeec1db30ba879a52a45c4ef0b5

zacharlie:
I managed to get the build running on Windows 10 21H2 pretty simply on the master branch. I think this is a local environment configuration issue. GDAL and Windows can have issues though. I default to using conda when I work with gdal. I used the following commands to set up a fresh instance:

```pwsh
git clone https://github.com/kartoza/docker-geoserver.git
cd docker-geoserver
conda create -n geoserver-build -c conda-forge python gdal
conda activate geoserver-build
```

Then I get the active JAVA_HOME path from powershell (you can use `(get-command java).Source`, but I figure the shortpath for java is safer because "Program Files" including whitespace has never really made sense).

```pwsh
(New-Object -ComObject Scripting.FileSystemObject).GetFile((get-command java).Source).ShortPath
```

[...]

PR #427, opened 2022-07-17
add windows build instructions to readme

zacharlie:
As discussed in #425

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@
        * [Reverse Proxy using NGINX](#reverse-proxy-using-nginx)
    * [Kubernetes (Helm Charts)](#kubernetes-helm-charts)
    * [Contributing to the image](#contributing-to-the-image)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -104,9 +104,48 @@ For some recent builds it is necessary to set the JAVA_PATH as well (e.g. Apache
...
+conda activate geoserver-build
+```
+
+Modify the `.env` with the appropriate environment variables. It is recommended that shortpaths (without whitespace) are used with forward slashes to prevent errors. You can get the current java command short path with powershell:
+
+```pwsh
+(New-Object -ComObject Scripting.FileSystemObject).GetFile((get-command java).Source).ShortPath
+```
+
+Running the above command should yield a path similar to `C:/PROGRA~1/Java/JDK-15~1.2/bin/java.exe`, which can be assigned to `JAVA_HOME` in the environment confoguration file.
```
NyakudyaA, line 140: configuration

PR #474, opened 2023-01-19
Use failover transport for cluster

pedroetb:
When ActiveMQ broker is not available at geoserver node (slave) startup, connection get disabled and require manual action to enable it again. Using failover transport allows to retry connection until broker is available.


NyakudyaA:
Thanks @pedroetb What is the current image tag you are using to run this?

pedroetb:
It's working at least with `v2.20.1`, `v2.20.2`, `v2.20.4` and `v2.21.0`, using `/settings` overwrite.

With newer versions (`> v2.21.0`), there are troubles when `entrypoint` tries to change ownership of read-only files at `/settings`, by the way. But editing `cluster.properties` manually shows that broker connection works too, even at `v2.22.0`.

NyakudyaA:
LGTM

PR #482, opened 2023-01-24
adding database schema to disk quota store postgres backend, it works…

Martenz:
adding database schema to disk quota store postgres backend, it works only with public right now, can be used later when geoserver will fix this because by default it creates the two needed tables only in default public schema

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -87,13 +87,6 @@ chown -R "${USER_NAME}":"${GEO_GROUP_NAME}" "${CATALINA_HOME}" "${FOOTPRINTS_DAT
 "${GEOSERVER_HOME}" "${EXTRA_CONFIG_DIR}"  /usr/share/fonts/ /scripts /tomcat_apps.zip \
 /tmp/ "${GEOWEBCACHE_CACHE_DIR}";chmod o+rw "${CERT_DIR}";chmod 400 ${CATALINA_HOME}/conf/*
 
-echo -e "[Entrypoint] Checking PostgreSQL connection to see if init tables are loaded: \033[0m"
-if [[  "$DB_BACKEND" =~ [Pp][Oo][Ss][Tt][Gg][Rr][Ee][Ss] ]];then
-  export PGPASSWORD="${POSTGRES_PASS}"
-  postgres_ready_status ${HOST} ${POSTGRES_PORT} ${POSTGRES_USER} $POSTGRES_DB
-  create_gwc_tile_tables ${HOST} ${POSTGRES_PORT} ${POSTGRES_USER} $POSTGRES_DB $POSTGRES_SCHEMA
-fi
-
 if [[ -f ${GEOSERVER_HOME}/start.jar ]]; then
   exec gosu ${USER_NAME} ${GEOSERVER_HOME}/bin/startup.sh
 else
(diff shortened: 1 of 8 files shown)


NyakudyaA (approved):
As a workaround for now, can we get the DDL for table creation if the schema is not public and we init it ourselves using psql. We can then remove this logic when the upstream fix is applied.

Something like
```
if [ "${POSTGRES_SCHEMA}" != 'public' ]; then
psql -d gis -p 5432 -U docker -h localhost -c "create table etc"
fi
```

Martenz:
> As a workaround for now, can we get the DDL for table creation if the schema is not public and we init it ourselves using psql. We can then remove this logic when the upstream fix is applied.
> 
> Something like
> 
> ```
> if [ "${POSTGRES_SCHEMA}" != 'public' ]; then
> psql -d gis -p 5432 -U docker -h localhost -c "create table etc"
> fi
> ```

It can be a solution, will try to test it in local

Martenz:
I added both the schema and tables creations once db is ready to patch. So specifying a new schema it will create everything and set properly the jdbc connection with currentSchema parameter.

Martenz:
I still have a doubt, If I start with a new empty geoserver-data-dir it sets properly the store but what if I want to do a restart and change through env variable from H2 to JDBC and viceversa or change the Postgres DB? I tried but once defined the files then I can not change them.. I have to delete them form the mounted volume and do docker-compose down then up. I want to avoid manual tasks. Maybe I'm missing some already built in script that do this ? Thanks

scripts/functions.sh
```diff
@@ -408,4 +408,19 @@ function postgres_ready_status() {
...
+
+function create_gwc_tile_tables(){
+  HOST="$1"
+  PORT="$2"
+  USER="$3"
+  DB="$4"
+  POSTGRES_SCHEMA="$5"
+  if [ ${POSTGRES_SCHEMA} != 'public' ]; then
+   psql -d "$DB" -p "$PORT" -U "$USER" -h "$HOST" -c "CREATE SCHEMA IF NOT EXISTS ${POSTGRES_SCHEMA}"
+   psql -d "$DB" -p "$PORT" -U "$USER" -h "$HOST" -c "CREATE TABLE IF NOT EXISTS ${POSTGRES_SCHEMA}.tileset(key character varying(320) COLLATE pg_catalog.\"default\" NOT NULL,layer_name character varying(128) COLLATE pg_catalog.\"default\",gridset_id character varying(32) COLLATE pg_catalog.\"default\",blob_format character varying(64) COLLATE pg_catalog.\"default\",parameters_id character varying(41) COLLATE pg_catalog.\"default\",bytes numeric(21,0) NOT NULL DEFAULT 0,CONSTRAINT tileset_pkey PRIMARY KEY (key))"
```
NyakudyaA: COLLATE pg_catalog.\"default\" is not necessary here, I think it defaults to this after creation in the DB

scripts/entrypoint.sh
```diff
@@ -91,6 +91,7 @@ echo -e "[Entrypoint] Checking PostgreSQL connection to see if init tables are l
 if [[  "$DB_BACKEND" =~ [Pp][Oo][Ss][Tt][Gg][Rr][Ee][Ss] ]];then
   export PGPASSWORD="${POSTGRES_PASS}"
   postgres_ready_status ${HOST} ${POSTGRES_PORT} ${POSTGRES_USER} $POSTGRES_DB
+  create_gwc_tile_tables ${HOST} ${POSTGRES_PORT} ${POSTGRES_USER} $POSTGRES_DB $POSTGRES_SCHEMA
 fi
```
NyakudyaA: Let's move all the logic from line 91-95 to https://github.com/kartoza/docker-geoserver/blob/develop/scripts/start.sh#L48.
This is where the disk quota logic is available

Martenz:
It works now, I guess everything is done

NyakudyaA:
LGTM after the documentation fixes

Martenz:
I added also the logics to set Direct Integration with Geoserver WMS to create gwc-gs.xml that should solve also the other issue opened.

Should I fix also the documentation then right? I can add the new env var to the readme, should I do something more?

NyakudyaA:
LGTM, will do a followup later to iron out some documentation etc

PR #485, opened 2023-02-01
Created Documents from readme

krishnaglodha:
I have used material by mkdocs https://squidfunk.github.io/mkdocs-material/ and uploaded build in `docs/` folder

PR #491, opened 2023-02-16
added docs from krishnaglodha

NyakudyaA:
supersedes https://github.com/kartoza/docker-geoserver/pull/485
Credits go to @krisaoe


krishnaglodha:
Hi @NyakudyaA , thanks for merging, should we host statics docs somewhere and link in in Readme ?

NyakudyaA:
Yes, I think so, Maybe we can host the docs with GitHub pages. What's your suggestion here

krishnaglodha:
I haven't hosted with GH pages yes.. Are we looking to connect it with kartoza.com domain ? Then we can host on Kartoza's hosting

NyakudyaA:
I will ask how to do it from Kartoza side and revert to you

NyakudyaA:
Hi @krishnaglodha Can you please share the generated markdown files (.md) so that we can use those for the docs. We are currently using the static HTML files that were generated but it will be difficult to maintain as we would have to update the HTML files and we can also automate the building docs with GitHub actions

PR #507, opened 2023-03-11
Allow to change context-root

tlvu:
Set env var `GEOSERVER_CONTEXT_ROOT=another-geoserver-context-root` and Geoserver will be accessible at https://hostname/another-geoserver-context-root/ instead of the default https://hostname/geoserver/.

I basically did a big find for `webapps/geoserver` and replace with `webapps/${GEOSERVER_CONTEXT_ROOT}` in all the scripts.

`GEOSERVER_CONTEXT_ROOT` default to `geoserver` if not set so this change is 100% backward compatible.

This is the fix to my problem described here https://stackoverflow.com/questions/75692310/is-it-possible-to-change-geoserver-context-root

I have a working docker image if you want to try this out: `pavics/geoserver:2.22.2-kartoza-build20230226-r4-allow-change-context-root`, built from this Dockerfile https://github.com/tlvu/docker-geoserver/blob/8023ada95a90e4c521c90025c6c134591acb5c59/Dockerfile.custom.  Docker image is used on our side in this PR https://github.com/bird-house/birdhouse-deploy-ouranos/pull/20

Without this fix, we see the following errors when the container starts:
```
cp: target '/usr/local/tomcat/webapps/ogc-geoserver/WEB-INF/lib/' is not a directory
Archive:  /community_plugins/colormap-plugin.zip
  inflating: /tmp/gs_plugin/gs-colormap-2.22-SNAPSHOT.jar
 Enabling colormap-plugin for GeoServer       
cp: cannot create regular file '/usr/local/tomcat/webapps/ogc-geoserver/WEB-INF/lib/': No such file or directory
cp: cannot create regular file '/usr/local/tomcat/webapps/ogc-geoserver/WEB-INF/lib/': No such file or directory
[Entrypoint] GENERATED tomcat  PASSWORD:  Oq2Ol5PYKKbIpr32tF 
 ----------------------------------------------------- 
  Run password encryption once for the first runtime 
find: ‘/usr/local/tomcat/webapps/ogc-geoserver/WEB-INF/lib/’: No such file or directory
Error: Could not find or load main class digest.sh
Caused by: java.lang.ClassNotFoundException: digest.sh
```

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -29,6 +29,17 @@ mkdir -p  "${GEOSERVER_DATA_DIR}" "${CERT_DIR}" "${FOOTPRINTS_DATA_DIR}" "${FONT
 source /scripts/functions.sh
 source /scripts/env-data.sh
 
+# Rename to match wanted context-root and so that we can unzip plugins to
+# existing directory.
+if [ x"${GEOSERVER_CONTEXT_ROOT}" != xgeoserver ]; then
+  echo "INFO: changing context-root to '${GEOSERVER_CONTEXT_ROOT}'."
+  if [ -e "${CATALINA_HOME}/webapps/geoserver" ]; then
+    mv "${CATALINA_HOME}/webapps/geoserver" "${CATALINA_HOME}/webapps/${GEOSERVER_CONTEXT_ROOT}"
+  else
+    echo "WARN: '${CATALINA_HOME}/webapps/geoserver' not found, probably already renamed as this is probably a container restart and not first run."
+  fi
+fi
+
 # Credits https://github.com/kartoza/docker-geoserver/pull/371
 set_vars
 export  READONLY CLUSTER_DURABILITY BROKER_URL EMBEDDED_BROKER TOGGLE_MASTER TOGGLE_SLAVE BROKER_URL
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -33,10 +33,11 @@ source /scripts/env-data.sh
 # existing directory.
 if [ x"${GEOSERVER_CONTEXT_ROOT}" != xgeoserver ]; then
   echo "INFO: changing context-root to '${GEOSERVER_CONTEXT_ROOT}'."
-  if [ -e "${CATALINA_HOME}/webapps/geoserver" ]; then
-    mv "${CATALINA_HOME}/webapps/geoserver" "${CATALINA_HOME}/webapps/${GEOSERVER_CONTEXT_ROOT}"
+  GEOSERVER_INSTALL_DIR="$(detect_install_dir)"
+  if [ -e "${GEOSERVER_INSTALL_DIR}/webapps/geoserver" ]; then
+    mv "${GEOSERVER_INSTALL_DIR}/webapps/geoserver" "${GEOSERVER_INSTALL_DIR}/webapps/${GEOSERVER_CONTEXT_ROOT}"
   else
-    echo "WARN: '${CATALINA_HOME}/webapps/geoserver' not found, probably already renamed as this is probably a container restart and not first run."
+    echo "WARN: '${GEOSERVER_INSTALL_DIR}/webapps/geoserver' not found, probably already renamed as this is probably a container restart and not first run."
   fi
 fi
 
--- a/scripts/functions.sh
+++ b/scripts/functions.sh
@@ -129,6 +129,14 @@ function validate_geo_install() {
 
 }
 
+function detect_install_dir() {
+  if [[ -f ${GEOSERVER_HOME}/start.jar ]]; then
+    return "${GEOSERVER_HOME}"
+  else
+    return "${CATALINA_HOME}"
+  fi
+}
+
 function unzip_geoserver() {
   if [[ -f /tmp/geoserver/geoserver.war ]]; then
     unzip /tmp/geoserver/geoserver.war -d "${CATALINA_HOME}"/webapps/${GEOSERVER_CONTEXT_ROOT} &&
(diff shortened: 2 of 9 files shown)


NyakudyaA:
Thanks @tlvu Will take  a look during the week

scripts/entrypoint.sh
```diff
@@ -29,6 +29,17 @@ mkdir -p  "${GEOSERVER_DATA_DIR}" "${CERT_DIR}" "${FOOTPRINTS_DATA_DIR}" "${FONT
 source /scripts/functions.sh
 source /scripts/env-data.sh
 
+# Rename to match wanted context-root and so that we can unzip plugins to
+# existing directory.
+if [ x"${GEOSERVER_CONTEXT_ROOT}" != xgeoserver ]; then
+  echo "INFO: changing context-root to '${GEOSERVER_CONTEXT_ROOT}'."
+  if [ -e "${CATALINA_HOME}/webapps/geoserver" ]; then
```
NyakudyaA: The webapp is not only deployed as `"${CATALINA_HOME}/webapps/geoserver"`, in some cases it can be defined as https://github.com/kartoza/docker-geoserver/blob/develop/scripts/setup.sh#L97

NyakudyaA:
@tlvu What about https://github.com/kartoza/docker-geoserver/blob/develop/build_data/index.jsp#L2, it seems like it's not handled

tlvu:
> @tlvu What about https://github.com/kartoza/docker-geoserver/blob/develop/build_data/index.jsp#L2, it seems like it's not handled

Oh wow, good catch.  How is that `index.jsp` being used?  The docker custom build I made to include my fix is running without any errors so how would I test the fix for that file?

scripts/entrypoint.sh
tlvu: I think maybe the case in https://github.com/kartoza/docker-geoserver/blob/a54266835d1ec330f0f2920ce28839732c554ccd/scripts/setup.sh#L97-L102 probably do not have to be handled?

At build time, we'll just keep the default "geoserver" name.  It's only at run time that we need to rename it to match required context-root.

Unless I missed something, `setup.sh` is used at build time so we can just keep it as-is.  `entrypoint.sh` is used at run time so it has to change.

In fact, at build time, we have to use the default "geoserver" name so that at run time we know for sure from name to rename to the required context-root.

tlvu:
> > @tlvu What about https://github.com/kartoza/docker-geoserver/blob/develop/build_data/index.jsp#L2, it seems like it's not handled
> 
> Oh wow, good catch. How is that `index.jsp` being used? The docker custom build I made to include my fix is running without any errors so how would I test the fix for that file?

Found it https://github.com/kartoza/docker-geoserver/blob/a54266835d1ec330f0f2920ce28839732c554ccd/scripts/start.sh#L250-L253

We do not use `ROOT_WEBAPP_REDIRECT` so we never caught this omission.  Will send a fix.

scripts/entrypoint.sh
NyakudyaA: Yes, the setup is run at build time but https://github.com/kartoza/docker-geoserver/blob/develop/scripts/entrypoint.sh#L102 
shows that if you install the platform-independent binary you can end up with your webapps as `/geoserver/webapps/geoserver`, so that would also need to be handled

tlvu:
> We do not use `ROOT_WEBAPP_REDIRECT` so we never caught this omission. Will send a fix.

Done ff75eb98ceccdd9880ec8b0fe6a7657c6260bae5

scripts/entrypoint.sh
tlvu: Oh thanks for pointing this out!  It seems we do not use this scenario so it never broke for us.  Will try to send a fix.

tlvu: I am confused, `GEOSERVER_HOME=/geoserver` so `${GEOSERVER_HOME}/start.jar` is `/geoserver/start.jar` but that file is not in the docker image.  So how do I get that file so I can test for this scenario?

tlvu: I think I finally understand what you meant.

There are 2 possible install dirs (CATALINA_HOME and GEOSERVER_HOME).  I already handed CATALINA_HOME, have to also handle GEOSERVER_HOME.

Hopefully this commit 6252ae3f3dc534d4bcf27a41ad0c11907ebfc936 is doing what you want.

scripts/functions.sh
```diff
@@ -246,11 +254,8 @@ function install_plugin() {
   if [[ -f "${DATA_PATH}"/"${EXT}".zip ]];then
      unzip "${DATA_PATH}"/"${EXT}".zip -d /tmp/gs_plugin
      echo -e "\e[32m Enabling ${EXT} for GeoServer \033[0m"
-     if [[ -f /geoserver/start.jar ]]; then
-       cp -r -u -p /tmp/gs_plugin/*.jar /geoserver/webapps/geoserver/WEB-INF/lib/
```
tlvu, line 250: I am not sure why `$GEOSERVER_HOME` is not used here but hardcode `/geoserver` is used.

If the hardcode flavor is needed because `$GEOSERVER_HOME` is not defined at that point, I can rollback the change here, let me know.  Or let me know how to get `$GEOSERVER_HOME` to be defined at that point.

tlvu:
This new `GEOSERVER_CONTEXT_ROOT` config should be documented in `README.md`.

Should you do it so the writing style is consistent or you want to let me do it?

scripts/functions.sh
NyakudyaA: $GEOSERVER_HOME is defined as an environment variable in the Dockerfile. So he hardcoded value is actually a mistake and should not be there

NyakudyaA:
You can add it and I will clean up later if needed in a separate PR

tlvu:
Documentation added in 96e4dd7157b9b308a718e6614e542b0c890ce562.

I found an error with my recent code change, fixed in c270f65dbc906e40cdd6d171aab112ecc9d7f8d1 

I have a working docker image if you want to try it `pavics/geoserver:2.22.2-kartoza-build20230226-r9-allow-change-context-root-and-fix-missing-stable-plugins-and-avoid-chown-datadir`

Also updated with latest of `develop` branch.  This PR is ready to be merged.

NyakudyaA:
thanks, @tlvu

pisaia01:
Hi,
using the following image
kartoza/geoserver:2.23.1

when I change the GEOSERVER_CONTEXT_ROOT nothing happens. I have a reverse proxy going from www.mydomain.com/customgeoserver to the geoserver port, 
and even if I change the GEOSERVER_CONTEXT_ROOT when I enter the URL I get redirected to www.mydomain.com/geoserver/index.html instead of customgeoserver/index.html.

Any updates?

PR #516, opened 2023-04-01
fixes #515

### LINKED ISSUE #515: Geoserver startup is significantly slower due to chown of GEOSERVER_DATA_DIR and GEOWEBCACHE_CACHE_DIR

This line https://github.com/kartoza/docker-geoserver/blob/b8d43882c674c3c1a9569d483f00e546e2895654/scripts/entrypoint.sh#L84-L88 makes Geoserver startup significantly slower for us because our `GEOSERVER_DATA_DIR` is 229 GB and `GEOWEBCACHE_CACHE_DIR` is under `GEOSERVER_DATA_DIR` which basically make it `chown` twice !

Found when upgrading from 2.19.0 to 2.22.2.  That new `chown` was introduced somewhere in 2.21.x I believe.

Please make it optional to `chown` `GEOSERVER_DATA_DIR` and `GEOWEBCACHE_CACHE_DIR`.  

Allow toggle env var like `CHOWN_DATA_DIR=false` and `CHOWN_CACHE_DIR=false`, like `EXISTING_DATA_DIR=false` would be great.

Have a gigantic `GEOSERVER_DATA_DIR`, see all Geoserver configs in our PR https://github.com/bird-house/birdhouse-deploy/pull/307.

### Versions

2.22.2

### Additional context

_No response_

NyakudyaA:
Added two env variables 
* CHOWN_DATA_DIR and CHOWN_GWC_DATA_DIR which default to true, this can be set to false to skip changing folder and file permission. The assumption is that a user should know the implications here as the startup will fail to create files in that directory without correct perms
* Added option to change perms only once for folders on startup unless the users change
* Added option to skip double chown of gwc and data dir is they are nested

### DIFF
--- a/build_data/stable_plugins.txt
+++ b/build_data/stable_plugins.txt
@@ -3,6 +3,7 @@ authkey-plugin
 cas-plugin
 charts-plugin
 css-plugin
+csw-iso-plugin
 db2-plugin
 dxf-plugin
 excel-plugin
@@ -18,6 +19,7 @@ importer-plugin
 jp2k-plugin
 mapml-plugin
 mbstyle-plugin
+metadata-plugin
 mongodb-plugin
 mysql-plugin
 netcdf-out-plugin
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -82,10 +82,21 @@ export JAVA_OPTS="${JAVA_OPTS} ${GEOSERVER_OPTS}"
 
 
 # Chown again - seems to fix issue with resolving all created directories
-chown -R "${USER_NAME}":"${GEO_GROUP_NAME}" "${CATALINA_HOME}" "${FOOTPRINTS_DATA_DIR}" "${GEOSERVER_DATA_DIR}" \
-"${CERT_DIR}" "${FONTS_DIR}"  /home/"${USER_NAME}"/ "${COMMUNITY_PLUGINS_DIR}" "${STABLE_PLUGINS_DIR}" \
-"${GEOSERVER_HOME}" "${EXTRA_CONFIG_DIR}"  /usr/share/fonts/ /scripts /tomcat_apps.zip \
-/tmp/ "${GEOWEBCACHE_CACHE_DIR}";chmod o+rw "${CERT_DIR}";chmod 400 ${CATALINA_HOME}/conf/*
+CHOWN_LOCKFILE=/scripts/.permission_file.lock
+if [[ $(stat -c '%U' ${CATALINA_HOME}) != "${USER_NAME}" ]] && [[ $(stat -c '%G' ${CATALINA_HOME}) != "${GEO_GROUP_NAME}" ]];then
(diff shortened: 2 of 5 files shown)


build_data/stable_plugins.txt
```diff
@@ -18,6 +19,7 @@ importer-plugin
 jp2k-plugin
 mapml-plugin
 mbstyle-plugin
+metadata-plugin
```
tlvu, line 22: Thanks for updating `stable_plugins.txt` as well.

scripts/entrypoint.sh
```diff
@@ -82,10 +82,21 @@ export JAVA_OPTS="${JAVA_OPTS} ${GEOSERVER_OPTS}"
...
+  if [[ ! -f ${CHOWN_LOCKFILE} ]];then
+    chown -R "${USER_NAME}":"${GEO_GROUP_NAME}" "${CATALINA_HOME}"   \
+    /home/"${USER_NAME}"/ "${COMMUNITY_PLUGINS_DIR}" "${STABLE_PLUGINS_DIR}" \
+    "${GEOSERVER_HOME}" /usr/share/fonts/ /scripts /tomcat_apps.zip /tmp/
+    touch ${CHOWN_LOCKFILE}
+  fi
+fi
+
+chown -R "${USER_NAME}":"${GEO_GROUP_NAME}"  "${FOOTPRINTS_DATA_DIR}" "${CERT_DIR}" "${FONTS_DIR}"  \
+   "${EXTRA_CONFIG_DIR}" ;chmod o+rw "${CERT_DIR}";gwc_file_perms ;chmod 400 ${CATALINA_HOME}/conf/*
```
tlvu, line 99: Just curious here, why not put this section in the same section that `touch ${CHOWN_LOCKFILE}`?

tlvu (commented):
Sorry for the very late review, I had too much on my plate.

The changes look good to me.

For us, we will be using the 2 toggles CHOWN_DATA_DIR and CHOWN_GWC_DATA_DIR.

Can't wait to try these on the next version of Geoserver.

tlvu:
Should these 2 new toogles be documented in the `README.md`?

NyakudyaA:
> Sorry for the very late review, I had too much on my plate.
> 
> 
> 
> The changes look good to me.
> 
> 
> 
> For us, we will be using the 2 toggles CHOWN_DATA_DIR and CHOWN_GWC_DATA_DIR.

> 
> 
> 
> Can't wait to try these on the next version of Geoserver.

If your data dir nests the gwc you will need to use only CHOWN_DATA_DIR otherwise both

NyakudyaA:
> Should these 2 new toogles be documented in the `README.md`?

Yes, I will raise a PR

scripts/entrypoint.sh
NyakudyaA: Because I have tried to identify static values and those that can change frequently. Example `scripts` folder will not be changed as it's used in the build but Footprints directory can be changed on restart of the container

PR #517, opened 2023-04-04
improve web.xml

NyakudyaA:
Based on tomcat configurations and https://github.com/bird-house/birdhouse-deploy/pull/307

PR #518, opened 2023-04-06
V2.23.0

NyakudyaA:
Upgrade to v2.23.0

### DIFF
--- a/build_data/web.xml
+++ b/build_data/web.xml
@@ -596,36 +596,25 @@
                so the entrypoint script can deactivate Cors settings if DISABLE_CORS is set -->
     <!-- CORS_START -->
     <filter>
-      <filter-name>CorsFilter</filter-name>
-      <filter-class>org.apache.catalina.filters.CorsFilter</filter-class>
-      <init-param>
-        <param-name>cors.allowed.origins</param-name>
-        <param-value>*</param-value>
-      </init-param>
-      <init-param>
-        <param-name>cors.allowed.methods</param-name>
-        <param-value>GET,POST,HEAD,OPTIONS,PUT</param-value>
-      </init-param>
-      <init-param>
-        <param-name>cors.allowed.headers</param-name>
-        <param-value>Content-Type,X-Requested-With,accept,Origin,Access-Control-Request-Method,Access-Control-Request-Headers,If-Modified-Since,Range,Authorization</param-value>
-      </init-param>
-      <init-param>
-        <param-name>cors.exposed.headers</param-name>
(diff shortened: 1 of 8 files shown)


build_data/web.xml
```diff
@@ -596,36 +596,25 @@
...
     <filter>
-      <filter-name>CorsFilter</filter-name>
-      <filter-class>org.apache.catalina.filters.CorsFilter</filter-class>
-      <init-param>
-        <param-name>cors.allowed.origins</param-name>
-        <param-value>*</param-value>
-      </init-param>
-      <init-param>
-        <param-name>cors.allowed.methods</param-name>
-        <param-value>GET,POST,HEAD,OPTIONS,PUT</param-value>
```
tlvu, line 607: Is there a reason why this is rollback from https://github.com/kartoza/docker-geoserver/pull/517?  It brokes some workflow?

NyakudyaA: This breaks the image, tried to build it locally and it was erroring with those changes.

NyakudyaA: @tlvu I get the following error when running with that XML
```
12-Apr-2023 18:40:11.018 SEVERE [main] org.apache.catalina.core.StandardContext.startInternal One or more Filters failed to start. Full details will be found in the appropriate container log file
12-Apr-2023 18:40:11.019 SEVERE [main] org.apache.catalina.core.StandardContext.startInternal Context [/geoserver] startup failed due to previous errors
12 Apr 18:40:11 CONFIG [gwc.layer] - GeoServer TileLayer store base directory is: gwc-layers
12 Apr 18:40:11 CONFIG [gwc.layer] - Loading tile layers from gwc-layers
12 Apr 18:40:11 CONFIG [gwc.layer] - Loaded 0 tile layers in 658.0 μs
12 Apr 18:40:11 INFO   [jdbc.JDBCQuotaStore] - Closing up the JDBC quota store
```

NyakudyaA: and the tests do not pass with those changes

tlvu: Wow I am surprised.  The param `cors.allowed.methods` was enabled for 2.19.0, for us, without any issues.

However we did not have `cors.support.credentials` and `cors.preflight.maxage` so they looked more suspect.

The rest of the changes, you simply changed the indentation so should not matter.

NyakudyaA: I will try another build without cors.preflight.maxage and see if it works. By the way web.xml is completely customizable as a run time config so this doesn't break the image

PR #525, opened 2023-04-29
Fixes #293 - although this is marked as resolved, the resolution isn't entirely correct

### LINKED ISSUE #293: Web Interface Login button/POST "http" while all other links "https"
Whenever I try to login to my geoserver with a browser, it is failing to connect.  I believe i've tracked it down to the fact that the login url is being presented to the client as HTTP, which is strange because all other URLs on the site are HTTPS as i expect, but not Login.

My system is setup behind a Load Balancer which handles certs/TLS termination, and then proxies requests to the docker container.  There's no expectation that the docker container will get an HTTPS request, or that letsencrypt will run or anything.  This seems to be fine for all web interface links except Login.

Is there an environment variable I can configure to tell the Login button to use `https://` without requiring the whole container to have self-managed certs?  Is this a bug with geoserver?

carderm:
This can cause the local healthcheck to fail.
This fix is for hosting Geoserver behind an SSL layer e.g. AWS Load Balancer or Cloudfront. It will  allow it to be accessed via http OR https.
1. nginx.conf: the proxy headers are passed by the Nginx reverse proxy
2. server.xml: a fix for tomcat to accept the proxy scheme (e.g. https) using RemoteIpValve
*There may be a better way to write the xml insertion..

### DIFF
--- a/scripts/functions.sh
+++ b/scripts/functions.sh
@@ -292,9 +292,6 @@ function activate_gwc_global_configs() {
       # default value
       envsubst < /build_data/gwc-gs.xml > "${GEOSERVER_DATA_DIR}"/gwc-gs.xml
     fi
-  else
-      # default value
-      envsubst < /build_data/gwc-gs.xml > "${GEOSERVER_DATA_DIR}"/gwc-gs.xml
   fi
 }
 
--- a/scripts/functions.sh
+++ b/scripts/functions.sh
@@ -276,7 +276,7 @@ function jdbc_disk_quota_config() {
   if [[ ! -f "${GEOWEBCACHE_CACHE_DIR}"/geowebcache-diskquota-jdbc.xml ]]; then
     # If it doesn't exists, copy from /settings directory if exists
     if [[ -f "${EXTRA_CONFIG_DIR}"/geowebcache-diskquota-jdbc.xml ]]; then
-      envsubst < "${EXTRA_CONFIG_DIR}"/geowebcache-diskquota-jdbc.xml < "${GEOWEBCACHE_CACHE_DIR}"/geowebcache-diskquota-jdbc.xml
+      envsubst < "${EXTRA_CONFIG_DIR}"/geowebcache-diskquota-jdbc.xml > "${GEOWEBCACHE_CACHE_DIR}"/geowebcache-diskquota-jdbc.xml
     else
       # default value
       envsubst < /build_data/geowebcache-diskquota-jdbc.xml > "${GEOWEBCACHE_CACHE_DIR}"/geowebcache-diskquota-jdbc.xml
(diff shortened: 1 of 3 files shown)


NyakudyaA:
@carderm Can you elaborate a bit on why the issue was not properly resolved? Adding the following env values does it for me when terminating in SSL
```
HTTP_SCHEME=https
HTTP_PROXY_NAME=foo.org
```

carderm:
> 

The issue is it *forces* any access method to be https. This can be an issue when also trying to access the server via http (e.g. for the local health check: [here](https://github.com/kartoza/docker-geoserver/blob/a433c2d16729a52dbd82ebfd52db67ac93a6579b/docker-compose.yml#L45)
Whereas this commit will allow the http/https scheme to be passed to Tomcat/Geoserver and it will render the URLS (e.g. login etc) as whichever scheme it's hosted as.

NyakudyaA:
@carderm Thanks for the clarification. So how can I test what this PR does vs the default behavior. If I do a health check when terminating in SSL should the health check fail in the other versions of the image

carderm:
well you can put a SSL Nginx server in front of it all and then access it via http (via your original Nginx) AND https (via the new one - see below) > you should see that both render correctly and you can login to either end point and Geoserver will remain as the correct scheme that you're viewing it as...

```
# Dockerfile
FROM nginx:alpine

# Create SSL directory
RUN mkdir -p /etc/nginx/ssl

# Install OpenSSL
RUN apk add --no-cache openssl

# Generate SSL certificate
RUN openssl req -new -newkey rsa:2048 -days 365 -nodes -x509 \
    -subj "/C=US/ST=CA/L=San Francisco/O=My Organization/OU=My Unit/CN=localhost" \
    -keyout /etc/nginx/ssl/localhost.key \
    -out /etc/nginx/ssl/localhost.crt

# Copy Nginx configuration
COPY nginx.conf /etc/nginx/nginx.conf

EXPOSE 443

CMD ["nginx", "-g", "daemon off;"]
```

```
#nginx.conf
worker_processes  1;

events {
    worker_connections  1024;
}


http {
    upstream nginx {
        server nginx:80;
    }

    server {
        listen       443 ssl;
        server_name  localhost;

        ssl_certificate      /etc/nginx/ssl/localhost.crt;
        ssl_certificate_key  /etc/nginx/ssl/localhost.key;

        location / {
            proxy_pass         http://nginx;
            proxy_redirect     off;
            proxy_set_header   Host $host;
            proxy_set_header   X-Real-IP $remote_addr;
            proxy_set_header   X-Forwarded-For $proxy_add_x_forwarded_for;
            proxy_set_header   X-Forwarded-Proto $scheme;
        }

        location /hello {
            return 200 "Hello, world!";
        }
    }
}
```

````yaml
  # CloudFront Dev
  cloudfront:
    build:
      context: ./_test_ssl
    ports:
      - 443:443

```

NyakudyaA:
@carderm Can you resolve the issues I mentioned and we can merge

carderm:
> @carderm Can you resolve the issues I mentioned and we can merge

I have replied above, am I missing something?

PR #553, opened 2023-08-09
create intermedieate directories for GEOSERVER_CONTEXT_ROOT

m4ci3k2:
To support multipart CONTEXT_ROOT (for example `/prefix/geoserver`) the directory up to the last past needs to be created before moving the files.


NyakudyaA (commented):
It would be nice to add a test for this i.e copy the login logic and add a new test

PR #565, opened 2023-09-15
Add ability to run as root

### LINKED ISSUE #528: backup-restore-plugin failing with permissions issues

Permissions issues are causing the backup-restore-plugin to fail.

This issue is not present in 2.17 so I assume it's caused by the permissions change made in 2.18 when we moved away from root ownership.

Enable the backup-restore-plugin via ENV var:

```
    environment:
             - COMMUNITY_EXTENSIONS=backup-restore-plugin
```

Click on Data->Backup & Restore.
Creating a backup shows warnings:
```
NO Exceptions Detected.

WARNING:Could not create directory.
WARNING:Could not create directory.
WARNING:Could not create directory.
```

and restoring the backup fails with errors:

```
 Error occurred while trying to move a Resource!
java.lang.IllegalStateException: Could not create directory.
```

### Versions

2.23.0

### Additional context

Not present in 2.17

NyakudyaA:
Keep the option to run as root, useful to use this in GeoNode where the container needs to run as root
Possibly fixes #528

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -96,16 +96,17 @@ export JAVA_OPTS="${JAVA_OPTS} ${GEOSERVER_OPTS}"
 
 
 # Chown again - seems to fix issue with resolving all created directories
-dir_ownership=(${CATALINA_HOME} /home/"${USER_NAME}"/ "${COMMUNITY_PLUGINS_DIR}"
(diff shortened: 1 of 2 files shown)


NyakudyaA:
@genebobmiller This will allow you to run as root like in previous images, You will need to use the env variable `RUN_AS_ROOT=TRUE`

PR #566, opened 2023-09-15
fix #563

### LINKED ISSUE #563: Cluster geoserver logs folder

Hello,

When using geoserver in clustered environnement with a shared data dir, the logs of each instance are written in log4j.appender.logfile.File = ${GEOSERVER_DATA_DIR}/geoserver.log as mentionned in build_data/log4j.properties
I think that each instance should use it's file for logs and we should use log4j.appender.logfile.File = ${CLUSTER_CONFIG_DIR}/geoserver.log

Regards,
Ahmed

Create a clustered environement of geoserver and start at least 2 instances

### Versions

2.23.1

### Additional context

_No response_

NyakudyaA:


NyakudyaA:
@ze-ahmed You can take a look at this

ze-ahmed:
I'm not sure it will fix the issue as each instance must has it's own log file and here your are using GEOSERVER_DATA_DIR which could be a shared folder in case we use shared directory between instances.

NyakudyaA:
> I'm not sure it will fix the issue as each instance must has it's own log file and here your are using GEOSERVER_DATA_DIR which could be a shared folder in case we use shared directory between instances.

This is fixed further in the follow up PR https://github.com/kartoza/docker-geoserver/blob/a4c31bf8db3d27fc39dc0d354f4e7b31c23ac4eb/scripts/functions.sh#L370-L407

Issue #562, opened 2023-09-01
Getting 404 trying to reach geoserver served under multipart GEOSERVER_CONTEXT_ROOT


aceenprsu:
### What is the bug or the crash?

I am getting a 404 trying to use the functionality introduced in https://github.com/kartoza/docker-geoserver/pull/553

### Steps to reproduce the issue

Below is a link to my fork where I copied the context test and added a multipart path to it.
https://github.com/aceenprsu/docker-geoserver/commit/2a6cc037c7c3475687d1a21333d3862063e60cc6
This is the output I get from trying to run the contexts test:
```
GEOSERVER_ADMIN_PASSWORD: myawesomegeoserver
GEOSERVER_ADMIN_USER: admin
test_rest_endpoints_accessible (test_login.TestGeoServerREST) ... FAIL

======================================================================
FAIL: test_rest_endpoints_accessible (test_login.TestGeoServerREST)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tests/test_login.py", line 28, in setUp
    self.assertEqual(response.status_code, 200)
AssertionError: 404 != 200

----------------------------------------------------------------------
Ran 1 test in 0.009s

FAILED (failures=1)
```

### Versions

2.23.2

NyakudyaA:
@m4ci3k2 can you please confirm if this is a bug or this logic only works for one folder not nested folders

m4ci3k2:
@NyakudyaA @aceenprsu sorry, actually I was wrong as the "problem" that the MR solved does not exist (but the `mkdir` is harmless of course). To install a tomcat application on a multipart path the WAR/folder has to be named `directory#name.war`, so for example to have geoserver under path `some-path` you can set `GEOSERVER_CONTEXT_ROOT=some-path#geoserver`. 

This is a built-in Tomcat function.

Sorry for the confusion.

PR #570, opened 2023-09-24
Update README and add test for nested context-root

aceenprsu:
This PR addresses https://github.com/kartoza/docker-geoserver/issues/562.
Apache Tomcat is used to serve geoserver. 
Apache Tomcat uses # to denote a nested context root. See https://octopus.com/blog/defining-tomcat-context-paths#conclusion

Nested context root is still one folder in Tomcat which makes https://github.com/kartoza/docker-geoserver/pull/553 an unneeded change for a nested context root.

### DIFF
--- a/.github/workflows/build-latest.yaml
+++ b/.github/workflows/build-latest.yaml
@@ -47,6 +47,7 @@ jobs:
           - login
           - stores
           - context
+          - contexts
           - disk-quota
     steps:
       - uses: actions/checkout@v3
--- a/README.md
+++ b/README.md
@@ -558,6 +558,14 @@ GEOSERVER_CONTEXT_ROOT=my-geoserver
 The example above will deploy Geoserver at https://host/my-geoserver instead of
 the default location at https://host/geoserver.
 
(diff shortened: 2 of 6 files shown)


NyakudyaA (approved):
Thanks @aceenprsu, For the tests we could have reused the same logic as in https://github.com/kartoza/docker-geoserver/blob/develop/scenario_tests/login/test.sh after line https://github.com/kartoza/docker-geoserver/blob/develop/scenario_tests/context/test.sh#L35C1-L59C69 we do the following


```
# Update context password
sed -i 's/foobar/foobar#geoserver/g' docker-compose.yml
# Bring the services up again
${VERSION} up -d geoserver

services=("geoserver")

for service in "${services[@]}"; do

  # Execute tests
  sleep 60
  echo "Execute test for $service"
  ${VERSION} exec -T $service /bin/bash /tests/test.sh

done

${VERSION} down -v
sed -i 's/foobar#geoserver/foobar/g' docker-compose.yml
```

PR #612, opened 2024-01-19
add hostname suffix to logs

robertd:
This PR adds hostname suffix to logs which would ease the clustering and siloed "multiple single instances" (see diagram) implementations.

image


NyakudyaA:
LGTM

PR #627, opened 2024-03-06
Update monitoring defaults and don't overwrite existing conf files

alexkuretz:
This PR is based on [this discussion](https://github.com/kartoza/docker-geoserver/discussions/624) and updates the default audit values to match the [Geoserver defaults](https://docs.geoserver.org/main/en/user/extensions/monitoring/audit.html), along with exposing those values to the user in the .env file. It also changes behavior to not overwrite existing config files. I also fixed a couple of minor tab spacing issues in the areas I was already working.


scripts/functions.sh
```diff
@@ -554,11 +555,12 @@ function entry_point_script {
...
 
-if [[ -f "${EXTRA_CONFIG_DIR}"/monitor.properties ]]; then
-      envsubst < "${EXTRA_CONFIG_DIR}"/monitor.properties > "${GEOSERVER_DATA_DIR}"/monitoring/monitor.properties
-else
+  if [[ -f "${EXTRA_CONFIG_DIR}"/monitor.properties ]]; then
+        envsubst < "${EXTRA_CONFIG_DIR}"/monitor.properties > "${GEOSERVER_DATA_DIR}"/monitoring/monitor.properties
+  else
 
-cat > "${GEOSERVER_DATA_DIR}"/monitoring/monitor.properties <<EOF
+  if [[ ! -f "${GEOSERVER_DATA_DIR}"/monitoring/monitor.properties ]]; then
```
NyakudyaA: @alexkuretz
If this monitoring file is setup, how do we make sure that the new variable that has been updated persist in the new file. I think we should revert back the logic to set/generate the file on each start

NyakudyaA (changes requested):
@alexkuretz
For the monitoring file, you seem to have a logic to create the file once, if I update my env variables and run an upgrade my new variables will not be picked up. I think we should revert the logic of generating the configs on each start

scripts/functions.sh
alexkuretz: I see, we have different use models, but I understand your point. I have reverted my changes that prevented overwriting the monitor.properties and logging.xml files. 

I have renamed some environment variables to better align with Geoserver documentation and defaults, and exposed them to the user by adding them to the .env file.

GEOSERVER_LOG_PROFILE is more accurate than GEOSERVER_LOG_LEVEL in this case.

Exposing GEOSERVER_LOG_DIR allows the user to customize log location. For example, we use an EFS share for GEOSERVER_DATA_DIR but local EBS volume for log files. Previously I have had to override the log location with an ADDITIONAL_JAVA_STARTUP_OPTION to accomplish this.

How do you feel about these changes? I suppose one concern would be backwards compatibility, in case a user had been utilizing the undocumented GEOSERVER_LOG_LEVEL. If you agree with the changes we could add a check for existence of that environment variable and use it.

scripts/env-data.sh
```diff
@@ -293,10 +293,13 @@ if [ -z "${PROXY_BASE_URL_PARAMETRIZATION}" ]; then
...
 fi
 
-if [ -z "${GEOSERVER_LOG_LEVEL}" ]; then
-    GEOSERVER_LOG_LEVEL=DEFAULT_LOGGING
+if [ -z "${GEOSERVER_LOG_PROFILE}" ]; then
+    GEOSERVER_LOG_PROFILE=DEFAULT_LOGGING
 fi
 
+if [ -z "${GEOSERVER_LOG_DIR}" ]; then
+    GEOSERVER_LOG_DIR=/opt/geoserver/data_dir/logs/geoserver.log
```
NyakudyaA: @alexkuretz isn't this supposed to be
GEOSERVER_LOG_DIR=${GEOSERVER_DATA_DIR}/logs

scripts/functions.sh
```diff
@@ -394,22 +394,21 @@ function setup_logging() {
...
-    if [[ ${CLUSTERING} =~ [Tt][Rr][Uu][Ee] ]]; then
-        export LOG_PATH=${CLUSTER_CONFIG_DIR}/geoserver-${HOSTNAME}.log
-      else
-        create_dir "${GEOSERVER_DATA_DIR}"/logs
-        export LOG_PATH=${GEOSERVER_DATA_DIR}/logs/geoserver-${HOSTNAME}.log
-    fi
+  if [[ ${CLUSTERING} =~ [Tt][Rr][Uu][Ee] ]]; then
+      export LOG_PATH=${CLUSTER_CONFIG_DIR}/geoserver-${HOSTNAME}.log
+    else
+      create_dir "${GEOSERVER_LOG_DIR}"
```
NyakudyaA, line 400: We need to add this env variable to the emtrypoint logic where we chown for the directory to be owned by the specific user

NyakudyaA (commented):
@alexkuretz left a couple of comments in the code.

Otherwise LGTM after you address the changes

scripts/env-data.sh
alexkuretz: Yes, good catch!

scripts/functions.sh
alexkuretz: Updated the existing chown to use GEOSERVER_LOG_DIR instead of GEOSERVER_DATA_DIR/logs

alexkuretz:
> @alexkuretz left a couple of comments in the code.
> 
> Otherwise LGTM after you address the changes

Good catches, thank you. Changes have been added.

PR #639, opened 2024-05-07
Use fonts-liberation, rather than side-loading mscorefonts

miceg:
The build scripts currently side-load `ttf-mscorefonts-installer` from an (official) Brazilian Debian mirror; which has problems:

* The Docker image appears to be based on Ubuntu Focal, and that package is for Debian.
* That mirror hostname is single-homed in Brazil, which isn't the fastest place to fetch stuff from out-of-continent – I get > 300 ms latency from Australia.
* Fetching and installing packages in this way is vulnerable to MITM attacks, as the package is fetched over HTTP and there are no signature checks (unlike using `apt`).
* There's no way for a user to accept or decline the EULA. [Ubuntu's version of the package gets explicit user consent](https://bugs.launchpad.net/ubuntu/+source/msttcorefonts/+bug/670629), but there's no way to provide this in a Docker context (either at build or run time).

To fix all of that, this replaces the fonts with less-encumbered equivalents in `fonts-liberation`. End-users can still sideload whatever fonts they like using the existing mechanisms (and that's OK).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,8 +20,8 @@ RUN set -eux; \
     apt-get update; \
     apt-get -y --no-install-recommends install \
         locales gnupg2 wget ca-certificates rpl pwgen software-properties-common  iputils-ping \
-        apt-transport-https curl gettext fonts-cantarell lmodern ttf-aenigma \
-        ttf-bitstream-vera ttf-mscorefonts-installer ttf-sjfonts tv-fonts libapr1-dev libssl-dev git \
+        apt-transport-https curl gettext fonts-cantarell fonts-liberation lmodern ttf-aenigma \
+        ttf-bitstream-vera ttf-sjfonts tv-fonts libapr1-dev libssl-dev git \
         wget zip unzip curl xsltproc certbot  cabextract gettext postgresql-client figlet gosu gdal-bin libgdal-java; \
       dpkg-divert --local --rename --add /sbin/initctl \
       && (echo "Yes, do as I say!" | apt-get remove --force-yes login) \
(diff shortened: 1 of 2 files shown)


miceg:
I see [Ubuntu has introduced a `msttcorefonts/accepted-mscorefonts-eula` flag to their package](https://git.launchpad.net/ubuntu/+source/msttcorefonts/tree/update-ms-fonts?h=ubuntu/lunar), which is not present in the Debian version, so my change will actually break these fonts.

Unfortunately, there's no way to accept this EULA in a Docker image build process, so I'm going to change this to use the `fonts-liberation` package instead.

NyakudyaA:
@miceg Thanks for picking this up, We generally have to install packages that have no licence terms etc, So users still have the option to download and install these at runtime

miceg:
No worries – the `mscorefonts` package's licensing is a little different to the others.

I had a bit more of a dig today, and [it looks like Microsoft asked Ubuntu to put in the prompt](https://bugs.launchpad.net/ubuntu/+source/msttcorefonts/+bug/670629/comments/17).

As for bundling in a Docker image, the `mscorefonts` Debian package [makes this comment about the EULA](https://salsa.debian.org/debian/msttcorefonts/-/blob/f50f0987c04d0e14412dc95d538a244c10f4c914/update-ms-fonts#L116-122):

> These fonts were provided by Microsoft "in the interest of cross-
> platform compatibility".  This is no longer the case, but they are
> still available from third parties.
>
> You are free to download these fonts and use them for your own use,
> but you may not redistribute them in modified form, including changes
> to the file name or packaging format.

Read into that as you will. 😄

PR #640, opened 2024-05-07
Don't remove the login package

miceg:
It looks like the `login` package was marked for removal in https://github.com/kartoza/docker-geoserver/pull/45, in an attempt to reduce the final image size.

However, the `login` package is present in the `tomcat:9.0.85-jdk17-temurin-focal` base image:

```
$ docker run --rm -it tomcat:9.0.85-jdk17-temurin-focal dpkg -l login
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name           Version                  Architecture Description
+++-==============-========================-============-=================================
ii  login          1:4.8.1-1ubuntu5.20.04.4 arm64        system login tools
```

While the running `geoserver` image will now report lower disk space usage, this doesn't actually have any effect on the image's true on-disk size – [it's still present in the base layers](https://docs.docker.com/storage/storagedriver/).

This PR just removes that command, because it has no effect.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,6 @@ RUN set -eux; \
         ttf-bitstream-vera ttf-sjfonts tv-fonts  libapr1-dev libssl-dev git \
         wget zip unzip curl xsltproc certbot  cabextract gettext postgresql-client figlet gosu gdal-bin libgdal-java; \
       dpkg-divert --local --rename --add /sbin/initctl \
-      && (echo "Yes, do as I say!" | apt-get remove --force-yes login) \
       && apt-get clean \
       && rm -rf /var/lib/apt/lists/*; \
       # verify that the binary works


NyakudyaA:
@miceg I thought the removal of the login package was for security reasons not the size reduction. Let's get the actions to run and see if there are any security implications on the login page

miceg:
It's not clear to me what security advantage there is to doing this, or how an attacker could leverage the `login` package in a way that they couldn't with other system packages. The only setuid binary in the package is `newgrp`, which has similar functionality to `gosu` (which is explicitly installed).

I had a look on GitHub, and the only `Dockerfile`s which remove this package appear to be forks or derivatives of this repository: https://github.com/search?q=%22apt-get+remove+--force-yes+login%22+path%3ADockerfile&type=code&p=4

I couldn't find any hits with `--allow-remove-essential`, which replaces `--force-yes`.

I acknowledge that in normal operation (as a container), the package shouldn't be used by anything. The bit that worries me is that it's removing an "essential" package – the required "Yes, do as I say!" confirmation should be seen as a warning signal. :)

PR #646, opened 2024-05-12
update compose files to use env variables

NyakudyaA:


scripts/env-data.sh
```diff
@@ -450,7 +423,8 @@ if [ -z "${MONITORING_BBOX_LOG_LEVEL}" ];then
 fi
 
 if [ -z "${ENTITY_RESOLUTION_ALLOWLIST}" ];then
-  ENTITY_RESOLUTION_ALLOWLIST="www.w3.org|schemas.opengis.net|www.opengis.net|inspire.ec.europa.eu/schemas"
+  #ENTITY_RESOLUTION_ALLOWLIST="www.w3.org|schemas.opengis.net|www.opengis.net|inspire.ec.europa.eu/schemas"
+  ENTITY_RESOLUTION_ALLOWLIST="*"
```
miceg: This shouldn't be disabled by default – it's a security risk.

Have a look at #651 – I suspect you may have hit the same issue as I did, and there's a better work-around there which doesn't introduce security problems.

**Edit:** I've just found #650, and that confirmed my suspicion 😄

scenario_tests/clustering/tests/test_clustering_master.py
```diff
@@ -84,6 +85,10 @@ def test_publish_store(self):
             geo.publish_featurestore(workspace='%s'
                                                % self.geo_workspace_name, store_name='%s' % self.store_name,
                                      pg_table='states')
+            copy("/usr/local/tomcat/data/styles/default_polygon.sld", "/usr/local/tomcat/data/styles/states.sld")
+            layer_sld_file = "/usr/local/tomcat/data/styles/states.sld"
+            geo.upload_style(path=r'%s' % layer_sld_file, workspace='%s' % self.geo_workspace_name)
```
miceg: I'd use `str(layer_sld_file)` instead of string formatting – it's much cleaner.

Issue #642, opened 2024-05-07
validate_url and download_extension request every URL twice, and don't re-use connections, making builds really slow


miceg:
### What is the bug or the crash?

`validate_url` and `download_extension` request every URL twice:

https://github.com/kartoza/docker-geoserver/blob/2d02b4df3e26df1cbb64b3cb4757e2f7aed597a2/scripts/functions.sh#L15-L16

https://github.com/kartoza/docker-geoserver/blob/939ee005d27fe123ce75c832e9706fe2825dc642/scripts/functions.sh#L129-L130

1. They first make a HTTP head request, and check the server response code, but in different ways:
   * `validate_url` fetches the URL with `wget -S --spider`
   * `download_extension` fetches the URL with `curl --head`
2. The URL is fetched with `wget` normally, and saved to a file.

As a shell script, there's no connection re-use between `wget`/`curl` calls, so this goes through a TCP (and TLS) handshake twice. That's 1 round-trip for HTTP or 3 round-trips for HTTPS... before you've even sent the HTTP request (another round-trip).

Many requests go to the same host (eg: `build.geoserver.org`, `sourceforge.net`), so could benefit by passing a URL list to `wget` rather than starting up new processes.

### Steps to reproduce the issue

Build the image with any significant amount of latency to `build.geoserver.org` or other sites – it takes many minutes.

### Versions

Current develop branch: https://github.com/kartoza/docker-geoserver/commit/2d02b4df3e26df1cbb64b3cb4757e2f7aed597a2

### Additional context

I'm sitting at about 200 ms latency to `build.geoserver.org`, so each HTTPS request takes at least 800ms (3 round-trips setup + 1 round-trip for the actual HTTP request) before the server has even sent a single byte of content. Doing that *twice* blows it out to 1.6 seconds/URL, and that's just network latency.

Looking at *just* the community plugins, with 68 community plugins, sending just *one* request per plugin would reduce build time by at least 54 seconds (at 200 ms latency).

If the download code was refactored so that all the plugin download requests happened in the same `wget` process, it could benefit from connection re-use, and get off 3 round-trips for every URL after the first (204 round-trips total = 41 seconds).

All up, fixing *just* the community plugins would reduce the build time at least 1.5 minutes (at 200 ms latency). I suspect the actual savings (for *all* downloads) would be much higher, and higher your latency, the bigger the impact.

[...]

PR #658, opened 2024-05-28
Don't re-download required plugins, update community plugin list.

miceg:
Some more prep work as part of #642:

* Create `required_plugins.txt` which lists [all mandatory plugins](https://github.com/kartoza/docker-geoserver/blob/7bcc741fdaef1c355e71e5c407b65ebf92cf96a5/scripts/setup.sh#L75-L79).

  A future PR will use this file as a download list.

* Remove all mandatory plugins from `stable_plugins.txt`, [which are already downloaded elsewhere](https://github.com/kartoza/docker-geoserver/blob/7bcc741fdaef1c355e71e5c407b65ebf92cf96a5/scripts/setup.sh#L75-L79).

* Add community plugin `wfs-freemarker-plugin`.

* Remove community plugin `rat-plugin` (doesn't exist anymore?).


aaime:
The rat community module [graduated to extension in 2.25.1](https://geoserver.org/announcements/vulnerability/2024/05/23/geoserver-2-25-1-released.html)

Issue #653, opened 2024-05-16
Broken WMS due to libjpeg-turbo errors in 2.25.0


vijdz:
### What is the bug or the crash?

WMS seems broken in 2.25.0.  
Layer preview doesn't always work, on a fresh container with sample raster data.

### Steps to reproduce the issue

Start a 2.25.0 container and try to preview any sample raster data, PNG or JPEG.  

The WMS returns the following xml for PNG preview of layer "nurc:Img_Sample" :  

```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?><!DOCTYPE ServiceExceptionReport SYSTEM "https://<hostname>/geoserver/schemas/wms/1.1.1/WMS_exception_1_1_1.dtd"> <ServiceExceptionReport version="1.1.1" >   <ServiceException>
      java.lang.RuntimeException: Error creating jpegturbo decompressor: No JPEG image is associated with this instance
</ServiceException></ServiceExceptionReport>
```

<details><summary>Here is the full stack trace for this PNG error</summary>

```
16 May 14:37:36 ERROR  [geoserver.ows] - 
java.lang.RuntimeException: Error creating jpegturbo decompressor: No JPEG image is associated with this instance
	at it.geosolutions.imageio.plugins.turbojpeg.TurboJpegImageReader.setInput(TurboJpegImageReader.java:199)
	at java.desktop/javax.imageio.ImageReader.setInput(ImageReader.java:380)
	at org.geotools.gce.image.WorldImageReader.getHRInfo(WorldImageReader.java:272)
	at org.geotools.gce.image.WorldImageReader.<init>(WorldImageReader.java:249)
	at org.geotools.gce.image.WorldImageFormat.getReader(WorldImageFormat.java:298)
	at org.geotools.gce.image.WorldImageFormat.getReader(WorldImageFormat.java:55)
	at org.geoserver.catalog.ResourcePool.getGridCoverageReader(ResourcePool.java:1689)
	at org.geoserver.catalog.ResourcePool.getGridCoverageReader(ResourcePool.java:1605)
	at org.geoserver.catalog.impl.CoverageInfoImpl.getGridCoverageReader(CoverageInfoImpl.java:174)
	at jdk.internal.reflect.GeneratedMethodAccessor231.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #659, opened 2024-05-28
Speed up plugin downloads

miceg:
**This PR is based on #658,** that'll need to be merged first to get a clean diff.

**Fixes:** #642 

This rearchitects the image's plugin and component downloads:

* Clean builds in my organisation's CI infrastructure went from 8 minutes to 4 minutes, and rebuilds take only 2 minutes.
* Clean builds in GitHub CI look to have dropped from around 15-20 minutes (and intermittently failing) to 7 minutes.

Detailed change list:

* Use `curl` rather than `wget`, because `wget1` doesn't support HTTP/2 and friends (which are faster).

* Download plugins using a separate Docker image, using a single `curl` process to benefit from connection re-use:

  * This runs in parallel with the main image – clean builds can download the plugins while installing dependencies from `apt`.
  * This benefits from Docker image caching on rebuilds – if the plugins list doesn't change, you don't have to download another 500MB of stuff.
  * This now propagates errors, rather than silently failing.

* Remove `validate_url` (which called `wget` twice), and just call `curl` directly.

* Make `download_extension` only try to download each extension once using `curl`, rather than twice using `curl` and `wget`.

* Simplify `package_geoserver` so it doesn't repeatedly repeat itself, and doesn't fetch each URL twice. Also makes it `exit 1` on errors.

* Clean up some strange shell syntax.

There are still some error handling weaknesses in the main image – the proper solution to that would be to use `set -e`. I'm hesitant to turn that on without further testing, because there may be assumptions in the existing script.

There are still some component downloads in the main image build which could be moved into the plugin downloader (such as GeoServer itself and `libjpeg-turbo`), but this is still a huge improvement.


build_data/plugin_download.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+set -eux
+
+mkdir -p /work/required_plugins
+mkdir -p /work/stable_plugins
+mkdir -p /work/community_plugins
+
+# Build a curl config to download all required plugins
+awk '{print "url = \"https://sourceforge.net/projects/geoserver/files/GeoServer/'"${GS_VERSION}"'/extensions/geoserver-'"${GS_VERSION}"'-"$0".zip\"\noutput = \"/work/required_plugins/"$0".zip\"\n--fail\n--location\n"}' < /work/required_plugins.txt > /work/curl.cfg
```
NyakudyaA: Can we not use the STABLE_PLUGIN_BASE_URL because sometimes the base url changes, I am not sure if it just redirects

build_data/plugin_download.sh
```diff
@@ -0,0 +1,22 @@
...
+
+if [ "${DOWNLOAD_ALL_STABLE_EXTENSIONS}" == "1" ]; then
+    # Add in all stable plugins
+    awk '{print "url = \"https://sourceforge.net/projects/geoserver/files/GeoServer/'"${GS_VERSION}"'/extensions/geoserver-'"${GS_VERSION}"'-"$0".zip\"\noutput = \"/work/stable_plugins/"$0".zip\"\n--fail\n--location\n"}' < /work/stable_plugins.txt >> /work/curl.cfg
```
NyakudyaA: Same as above

build_data/plugin_download.sh
```diff
@@ -0,0 +1,22 @@
...
+fi
+
+if [ "${DOWNLOAD_ALL_COMMUNITY_EXTENSIONS}" == "1" ]; then
```
NyakudyaA: What happens if my env eq 0, so nothing will be downloaded. The previous logic downloaded a single extention so that you can quickly test your build. Either we ditch the condition and arg

Dockerfile
```diff
@@ -1,18 +1,51 @@
...
+FROM --platform=$BUILDPLATFORM alpine:3.19 AS geoserver-plugin-downloader
+ARG GS_VERSION=2.25.0
+ARG DOWNLOAD_ALL_STABLE_EXTENSIONS=1
+ARG DOWNLOAD_ALL_COMMUNITY_EXTENSIONS=1
+
+RUN apk update && apk add curl
+
+WORKDIR /work
+ADD \
+    build_data/required_plugins.txt \
```
NyakudyaA, line 32: Since we are still running and installing these as part of the setup script could we rather have a new env MANDATORY_STABLE_EXTENSIONS=gdal-plugin, wps-plugin etc

We can then activate the mandatory plugins in the start.sh at the beginning.

This will give us control on which plugins to activate, if unset a user would then use the stable_plugins env to activate the ones he wants.

Maybe we could even remove the required_plugins.tx and just have all stable extension downloaded and the env would cherry pick the stable env so that they are installed as default

NyakudyaA (changes requested):
Nice, made a couple of suggestions

build_data/plugin_download.sh
miceg: Fixed

build_data/plugin_download.sh
miceg: Fixed

build_data/plugin_download.sh
miceg: I didn't consider this – that's strange behaviour. I guess this is why it is `DOWNLOAD_ALL_STABLE_EXTENSIONS` and not `DOWNLOAD_STABLE_EXTENSIONS`. 😅 

I can see value in being able to build with only "required" or "required + stable" plugins to get a smaller image; but I don't think this is built as a proper target anywhere.

I feel like only downloading the first plugin alphabetically in each category isn't really useful on its own – you end up with a strangely broken image.

Would you be happy with this becoming `DOWNLOAD_STABLE_EXTENSIONS` and `DOWNLOAD_COMMUNITY_EXTENSIONS` (ie: minus the `_ALL`) instead?

Dockerfile
miceg: > Maybe we could even remove the required_plugins.tx and just have all stable extension downloaded and the env would cherry pick the stable env so that they are installed as default

Downloading everything as `stable_extensions` [removes the option to make a smaller image](https://github.com/kartoza/docker-geoserver/pull/659/files#r1620076602), so I'm slightly less of a fan of that.

> Since we are still running and installing these as part of the setup script could we rather have a new env MANDATORY_STABLE_EXTENSIONS=gdal-plugin, wps-plugin etc

Yeah, I think it'd be good to be able to make the "required" set of extensions configurable – so that they're enabled by default, but you can still turn them off (eg: for mitigating things like #653).

But this would mean changing the `STABLE_EXTENSIONS` parameter – users would need to include the "required" set of extensions in here too, and all of those will need to be extracted at start-up time.

Maybe change this for a new `ACTIVE_EXTENSIONS` variable:

* `DEFAULT_EXTENSIONS` is an internal-only variable which contains a set of extensions as listed in `required_plugins.txt` (ie: not exposed as a parameter)
* `STABLE_EXTENSIONS` works as it does now
* `ACTIVE_EXTENSIONS` defaults to `DEFAULT_EXTENSIONS + STABLE_EXTENSIONS`

Then: 

* If someone wants to disable one/many/all of the "default" extensions, they override `ACTIVE_EXTENSIONS`
* If someone wants to add some stable extensions and keep the "default" extension set, they can add it to `STABLE_EXTENSIONS` (as they can now)
* If someone wants to disable one/many/all of the "default" extensions and add more stable extensions, they override `ACTIVE_EXTENSIONS`

This wouldn't impact community extensions, these would remain entirely separate.

What do you think?

build_data/plugin_download.sh
NyakudyaA: I am happy with changing it, The arg was only introduced to speed up the build of the image and use it in test with only a subset of the extensions. I am happy for it to be renamed, I don't think there are a lot of people who use it anyway when building their image

Dockerfile
NyakudyaA: I think that makes sense a lot

NyakudyaA:
@miceg Are you able to complete this soon? I need to publish the latest bug fix .1 version. If not I will push the new version and we can update this when you get to it

build_data/plugin_download.sh
NyakudyaA: removed this arg, if the download is improved, no need to have the arg, simply always download all plugins

Dockerfile
NyakudyaA: done in https://github.com/kartoza/docker-geoserver/pull/659/commits/0dedab4df3635203c09af7c70baa650e199521a5,

NyakudyaA (approved):

miceg:
Thanks for finishing that off – I've been stuck in some other work. The changes you made in the final version look good to me. :)

PR #699, opened 2024-11-05
Fixes for jdbcstore, jdbcconfig & jndi

tldev-de:
Hi there,

this is my first contribution to this project, so please give me feedback if there is anything I can do better.

This commit fixes two problems, I experienced while using it for a larger deployment:

1. jdbcstore does not get activated, since the changed configuration is written in `jdbcstore.propertiesproperties` instead of `jdbcstore.properties`
2. jdbcconfig & jdbcstore cannot be used with jndi enabled, since the postgresql driver is not available

The first issue is solved by just fixing the filename. The second issue is fixed by adding the respective line to enable jndi in [jdbcconfig](https://docs.geoserver.org/main/en/user/community/jdbcconfig/configuration.html#jndi) / [jdbcstore](https://docs.geoserver.org/2.22.x/en/user/community/jdbcstore/configuration.html#jndi).

if you need more details to understand these problems, please let me know.

Hope this helps :)


scripts/functions.sh
```diff
@@ -556,6 +556,13 @@ function setup_jdbc_db_config() {
             else
               envsubst < /build_data/jdbcconfig/jdbcconfig.properties > "${GEOSERVER_DATA_DIR}"/jdbcconfig/jdbcconfig.properties
             fi
+
+            # Set jndiName if POSTGRES_JNDI is set to true
+            sed -i '/^jndiName=/d' "${GEOSERVER_DATA_DIR}"/jdbcconfig/jdbcconfig.properties
```
NyakudyaA: shouldn't this line be inside the if statement?

scripts/functions.sh
```diff
@@ -586,9 +593,15 @@ function setup_jdbc_db_store() {
...
             if [[ -f "${EXTRA_CONFIG_DIR}"/jdbcstore.properties ]]; then
-              envsubst < "${EXTRA_CONFIG_DIR}"/jdbcstore.properties > "${GEOSERVER_DATA_DIR}"/jdbcstore/jdbcstore.propertiesproperties
+              envsubst < "${EXTRA_CONFIG_DIR}"/jdbcstore.properties > "${GEOSERVER_DATA_DIR}"/jdbcstore/jdbcstore.properties
             else
-              envsubst < /build_data/jdbcstore/jdbcstore.properties > "${GEOSERVER_DATA_DIR}"/jdbcstore/jdbcstore.propertiesproperties
+              envsubst < /build_data/jdbcstore/jdbcstore.properties > "${GEOSERVER_DATA_DIR}"/jdbcstore/jdbcstore.properties
+            fi
+
+            # Set jndiName if POSTGRES_JNDI is set to true
+            sed -i '/^jndiName=/d' "${GEOSERVER_DATA_DIR}"/jdbcstore/jdbcstore.properties
```
NyakudyaA: same as this as well

NyakudyaA (changes requested):
@tldev-de can we test this change, maybe provide a sample docker-compose with these settings and I will make a test in a separate PR

scripts/functions.sh
NyakudyaA: moved this to be used after setting default .properties

NyakudyaA (approved):

tldev-de:
Sorry for the delay - I was really busy the last few days. I actually deliberately left the `sed` command outside the if condition, because that makes it possible to switch between jndi and direct connection by just changing the ENV-variable. But of course the comment makes more sense inside the if condition. I can change that again if you like. :)

Regarding the docker compose file - this should be a good example:

```
volumes:
  geoserver-data-dir:
  postgis-data:


services:
  db:
    image: kartoza/postgis:16-3.4
    volumes:
      - postgis-data:/var/lib/postgresql
    environment:
      - POSTGRES_DB=gis
      - POSTGRES_USER=docker
      - POSTGRES_PASS=docker
      - ALLOW_IP_RANGE=0.0.0.0/0
      - POSTGRES_MULTIPLE_EXTENSIONS=postgis,hstore,postgis_topology,postgis_raster,pgrouting
      - RUN_AS_ROOT=true
    restart: on-failure
    healthcheck:
      test: "PGPASSWORD=docker pg_isready -h 127.0.0.1 -U docker -d gis"

  geoserver:
    image: 'kartoza/geoserver:2.26.0'
    restart: 'always'
    volumes:
      - geoserver-data-dir:/opt/geoserver/data_dir
    environment:
      GEOSERVER_ADMIN_PASSWORD: myawesomegeoserver
      GEOSERVER_ADMIN_USER: admin
      INITIAL_MEMORY: 2G
      MAXIMUM_MEMORY: 4G
      POSTGRES_JNDI: TRUE # this is important to enable jndi
      HOST: db
      POSTGRES_DB: gis
      POSTGRES_USER: docker
      POSTGRES_PASS: docker
      COMMUNITY_EXTENSIONS: "jdbcconfig-plugin, jdbcstore-plugin" # this is important to enable the plugins
    depends_on:
      db:
        condition: service_healthy
    ports:
      - "8080:8080"
    healthcheck:
      test: ["CMD-SHELL", "curl --fail --silent --write-out 'HTTP CODE : %{http_code}\n' --output /dev/null -u $${GEOSERVER_ADMIN_USER}:$${GEOSERVER_ADMIN_PASSWORD} http://localhost:8080/geoserver/rest/about/version.xml"]
      interval: 1m30s
      timeout: 10s
      retries: 3
```

NyakudyaA:
@tldev-de if you know it's better to have it outside then move it there. Also note that I moved the logic to be after setting the default configs. The extra configs already catered for this as you could add any king of configuration you could.

PR #729, opened 2025-04-15
Upgrade Geoserver version to 2.27.0

alexkuretz:
Upgraded Geoserver to 2.27.0.

Updated README to include updates to jdk base image in 2.26.2 release.

Added variables for DATA_DIR_LOADER options, not sure if I need to add anything to entrypoint.sh or if it's ok to have users modify those values in `ADDITIONAL_JAVA_STARTUP_OPTIONS` if they need to, thoughts?

The upgrade script wasn't updating the workflows, I tweaked the sed command and it worked for me, let me know if I misunderstood something about that.


README.md
```diff
@@ -100,20 +100,20 @@ See the [dockerhub tomcat](https://hub.docker.com/_/tomcat/)
...
 ```
-docker build --build-arg IMAGE_VERSION=9.0.91-jdk11-temurin-focal --build-arg JAVA_HOME=/usr/local/openjdk-11/bin/java --build-arg GS_VERSION=2.26.2 -t kartoza/geoserver:2.26.2 .
+docker build --build-arg IMAGE_VERSION=9.0.99-jdk11-temurin-noble --build-arg JAVA_HOME=/usr/local/openjdk-11/bin/java --build-arg GS_VERSION=2.27.0 -t kartoza/geoserver:2.27.0 .
 ```
 
 **Note:** Please check the [GeoServer documentation](https://docs.geoserver.org/stable/en/user/production/index.html) 
 to see which Tomcat versions are supported.
 
-The current build uses the base image `tomcat:9.0.91-jdk11-temurin-focal` because of the dependency on
+The current build uses the base image `tomcat:9.0.99-jdk11-temurin-noble` because of the dependency on
```
NyakudyaA: The whole logic about dependency on java etc can be scrapped as we are now inheriting the java bindings from upstream

scripts/env-data.sh
```diff
@@ -452,3 +452,12 @@ fi
 if [ -z ${GEOSERVER_FILESYSTEM_SANDBOX} ];then
    GEOSERVER_FILESYSTEM_SANDBOX=''
 fi
+
+# https://docs.geoserver.org/2.27.x/en/user/datadirectory/setting.html#configuration
+if [ -z ${GEOSERVER_DATA_DIR_LOADER_ENABLED} ];then
```
NyakudyaA, line 457: If `GEOSERVER_DATA_DIR_LOADER_ENABLED` and `GEOSERVER_DATA_DIR_LOADER_THREADS` have defaults which are recommended, let's add it to the entrypoint script

NyakudyaA (changes requested):
LGTM after addressing the comments, Also update the casing with

```
FROM ghcr.io/osgeo/gdal:ubuntu-full-${GDAL_VERSION} as gdal-builder
```
to become

```
FROM ghcr.io/osgeo/gdal:ubuntu-full-${GDAL_VERSION} AS gdal-builder
```

scripts/env-data.sh
alexkuretz: There are defaults applied by geoserver if these vars aren't supplied, "True" and 

> By default, the loader uses a heuristic that selects the minimum between 16 and the number of available processors as reported by the JVM

I was reluctant to set defaults since the heuristic is dynamic.

alexkuretz:
FIxed the Dockerfile syntax issue, and left comment on default values. I'll leave the README rewrite to you if you don't mind since I'm not an expert there, just saw an inconsistency to fix.

scripts/env-data.sh
NyakudyaA: I think we can set GEOSERVER_DATA_DIR_LOADER_ENABLED in the entry point by default while the other one might be difficult to set unless we can automatically detect how many threads are available on the jvm and set it to that. If not we can let the user pass it as additional jvm

README.md
NyakudyaA: @alexkuretz you can remove the whole statement from because and it should be fine

alexkuretz: Done

scripts/env-data.sh
alexkuretz: Done

alexkuretz:
Hi, what's the process to get these releases built and images published so they are available to use? I see there's been a few releases but no updated images since January.

NyakudyaA:
@alexkuretz I am working on a fix for https://github.com/kartoza/docker-geoserver/issues/732, I will have it ready in the next day or so

PR #741, opened 2025-05-15
fix: stop overriding `logging.xml` with default logging level

### LINKED ISSUE #739: `logging.xml` being overwritten with default values with every startup

I'm running kartoza/geoserver with a mounted pre created `data_dir`. I have [custom logging profile](https://docs.geoserver.org/2.22.x/en/user/configuration/logging.html#custom-logging-profiles) (json).
The default logging profile is being saved under `logging.xml`.
Each time my geoserver boots, `logging.xml` file is being overwritten with the default `logging.xml` file.

1. [Create custom logging profile](https://docs.geoserver.org/2.22.x/en/user/configuration/logging.html#custom-logging-profiles).
2. Run kartoza/geoserver:2.27.0 with mounted empty directory to `/opt/geoserver/data_dir`
3. Settings -> Global -> Logging profile -> choose any.
4. Delete the container and run it again. 
5. See `logging.xml` reverted back to default.

### Versions

2.27.0

### Additional context

_No response_

NivGreenstein:
Closes: #739 

According to my recently opened issue #739, I've decided to try and fix it myself. 
This issue also can be found with a leftover todo in `/scripts/start.sh`
```
# TODO: If this value is set and a user resets it in the UI, we need to respect that and not reset it again
export GEOSERVER_LOG_PROFILE
geoserver_logging
```

This PR introduce a fix to the same problem. 
* If `GEOSERVER_LOG_PROFILE` env is set -> its value will always be used. 
* If `GEOSERVER_LOG_PROFILE` isn't set and `logging.xml` exists,  the value from the file will be used. 
  * Most cases (like container reset), `DEFAULT_LOGGING` will be found. 
* Otherwise, `GEOSERVER_LOG_PROFILE` will default to `DEFAULT_LOGGING`.


scripts/functions.sh
```diff
@@ -359,13 +359,17 @@ function geoserver_logging() {
       export LOG_PATH=${GEOSERVER_LOG_DIR}/geoserver.log
   fi
 
-    echo "
+  if [ -f "${GEOSERVER_DATA_DIR}/logging.xml" ]; then
```
NyakudyaA: What if it already exists but a user want's to also change it's values from DEFAULT etc using the env variable.

NyakudyaA (commented):
I had come up with an alternate solution to this, haven't tested this as well
```
function set_logging_xml() {
       echo "
<logging>
  <level>${GEOSERVER_LOG_PROFILE}</level>
  <location>${LOG_PATH}</location>
  <stdOutLogging>true</stdOutLogging>
</logging>
" > "${GEOSERVER_DATA_DIR}"/logging.xml
  }
function geoserver_logging() {
  if [[ ${CLUSTERING} =~ [Tt][Rr][Uu][Ee] ]]; then
      export LOG_PATH=${CLUSTER_CONFIG_DIR}/geoserver-${HOSTNAME}.log
    else
      create_dir "${GEOSERVER_LOG_DIR}"
      export LOG_PATH=${GEOSERVER_LOG_DIR}/geoserver.log
  fi

  SETUP_LOG_FILE="${GEOSERVER_DATA_DIR}"/logging.xml

  if [[ -f $SETUP_LOG_FILE} ]];then
   # grab the current value of the log level set in the file
    LEVEL_VALUE=$(grep -oP '<level>\K[^<]+' "$SETUP_LOG_FILE" 2>/dev/null)
    if [[ ${GEOSERVER_LOG_PROFILE} != ${LEVEL_VALUE} ]];then
      export GEOSERVER_LOG_PROFILE=${LEVEL_VALUE}
      set_logging_xml
    else
        set_logging_xml
    fi
  else
    set_logging_xml
  fi
  

  if [[ ! -f ${LOG_PATH} ]];then
    touch "${LOG_PATH}"
  fi
}
```
What's your take ?

NivGreenstein:
> I had come up with an alternate solution to this, haven't tested this as well
> 
> ```
> function set_logging_xml() {
>        echo "
> <logging>
>   <level>${GEOSERVER_LOG_PROFILE}</level>
>   <location>${LOG_PATH}</location>
>   <stdOutLogging>true</stdOutLogging>
> </logging>
> " > "${GEOSERVER_DATA_DIR}"/logging.xml
>   }
> function geoserver_logging() {
>   if [[ ${CLUSTERING} =~ [Tt][Rr][Uu][Ee] ]]; then
>       export LOG_PATH=${CLUSTER_CONFIG_DIR}/geoserver-${HOSTNAME}.log
>     else
>       create_dir "${GEOSERVER_LOG_DIR}"
>       export LOG_PATH=${GEOSERVER_LOG_DIR}/geoserver.log
>   fi
> 
>   SETUP_LOG_FILE="${GEOSERVER_DATA_DIR}"/logging.xml
> 
>   if [[ -f $SETUP_LOG_FILE} ]];then
>    # grab the current value of the log level set in the file
>     LEVEL_VALUE=$(grep -oP '<level>\K[^<]+' "$SETUP_LOG_FILE" 2>/dev/null)
>     if [[ ${GEOSERVER_LOG_PROFILE} != ${LEVEL_VALUE} ]];then
>       export GEOSERVER_LOG_PROFILE=${LEVEL_VALUE}
>       set_logging_xml
>     else
>         set_logging_xml
>     fi
>   else
>     set_logging_xml
>   fi
>   
> 
>   if [[ ! -f ${LOG_PATH} ]];then
>     touch "${LOG_PATH}"
>   fi
> }
> ```
> 
> What's your take ?

Looks (haven't tested) same as my suggestion. Instead of replacing `<location>...</location>` using `sed` command, you still overwrite the file. But, this suggestion still won't resolve your comment. 

Note that you can reduce `set_logging_xml` to one call. 
```
if [[ -f $SETUP_LOG_FILE} ]];then
   # grab the current value of the log level set in the file
    LEVEL_VALUE=$(grep -oP '<level>\K[^<]+' "$SETUP_LOG_FILE" 2>/dev/null)
    if [[ ${GEOSERVER_LOG_PROFILE} != ${LEVEL_VALUE} ]];then
      export GEOSERVER_LOG_PROFILE=${LEVEL_VALUE}
    fi  
  fi
# Write to logging.xml
set_logging_xml
```

scripts/functions.sh
NivGreenstein: Currently, the default value for `GEOSERVER_LOG_PROFILE` is `DEFAULT_LOGGING`.
I can update the behavior so that if a `logging.xml` file already exists and the `GEOSERVER_LOG_PROFILE` environment variable is set to `DEFAULT_LOGGING`, the original logging.xml file will be preserved.
However, it is up to the maintainers to decide which value should take precedence.

NivGreenstein:
```suggestion
  if [ -f "${GEOSERVER_DATA_DIR}/logging.xml" ] && [ "${GEOSERVER_LOG_PROFILE}" = "DEFAULT_LOGGING" ]; then
```

What do you think?

NyakudyaA (changes requested):
I think let's add an env variable like
`OVERIDE_LOGGING=TRUE` which should cater for backward compatibility so that a user can control having this logic always reset to the value of the env value. This should also cover your use case when the file exist and you simply need to have the env set to false

NivGreenstein:
> I think let's add an env variable like `OVERIDE_LOGGING=TRUE` which should cater for backward compatibility so that a user can control having this logic always reset to the value of the env value. This should also cover your use case when the file exist and you simply need to have the env set to false

After considering your comment, I decided to go with your first suggested approach:
* If `GEOSERVER_LOG_PROFILE` env is set -> its value will always be used. 
* If `GEOSERVER_LOG_PROFILE` isn't set and `logging.xml` exists,  the value from the file will be used. 
  * Most cases (like container reset), `DEFAULT_LOGGING` will be found. 
* Otherwise, `GEOSERVER_LOG_PROFILE` will default to `DEFAULT_LOGGING`. 

This approach avoids introducing a new environment variable and ensures backward compatibility, without any breaking changes for existing users.

NivGreenstein:
@NyakudyaA Just pushed test cases to verify logic indeed works as expected

NyakudyaA:
Thanks for adding test, running the work flows and if it succeeds will approve

NyakudyaA (approved):

NivGreenstein:
@NyakudyaA Thanks for approving! 
Can you go ahead and close this PR? 
Waiting to use this new version 😁

PR #745, opened 2025-05-20
Add option to configure multiple users and roles

NyakudyaA:
TODO
* update readme 

Add an option to pass, i.e
```
GEOSERVER_ADMIN_PASSWORD=myawesomegeoserver,mygeoserver,mysample
GEOSERVER_ADMIN_USER=foo,myadmin,sample
```
If there is a mismatch on number of users and password, the default creds are used to login
Currently there is no logic to parse passwords with comma separated i.e 
```
GEOSERVER_ADMIN_PASSWORD="'myawes,omegeoserver',mygeoserver,mysample"

### DIFF
--- a/scenario_tests/login/docker-compose.yml
+++ b/scenario_tests/login/docker-compose.yml
@@ -62,3 +62,23 @@ services:
       timeout: 10s
       retries: 3
 
+  users:
+    image: 'kartoza/geoserver:${TAG:-manual-build}'
+    restart: 'always'
+    volumes:
+      - ./tests:/tests
+    environment:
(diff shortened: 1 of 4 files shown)


NyakudyaA:
@NivGreenstein @tharanathkartoza would you be keen to test this before I merge, i am thinking a practical use of this is mostly in GeoNode

NivGreenstein:
> @NivGreenstein @tharanathkartoza would you be keen to test this before I merge, i am thinking a practical use of this is mostly in GeoNode 

Hey! Thanks for tagging me. Although I don't really manage users in Geoserver's admin webapp, I think I have a test-case for it. 
I'll check it out next work-week.

tharanathkartoza:
sure @NyakudyaA .. will test with our new geonode image this week..

Issue #790, opened 2026-02-16
update_passwords logic error prevents using Docker Secrets


vchaparro:
### What is the bug or the crash?

When deploying kartoza/geoserver using Docker Secrets (GEOSERVER_ADMIN_PASSWORD_FILE), the container fails to update the admin password and falls back to default credentials. The startup script throws an error because it attempts to use a randomly generated password and the file-based secret simultaneously.

The error:
`error: both GEOSERVER_ADMIN_PASSWORD and GEOSERVER_ADMIN_PASSWORD_FILE are set (but are exclusive)`

This happens even when it's not set GEOSERVER_ADMIN_PASSWORD in the environment variables.

### Steps to reproduce the issue

1.  Create a Docker Swarm service or Docker Compose stack using kartoza/geoserver:2.27.1.
2. Define a Docker Secret for the admin password (e.g., geoserver_admin_password).
3.  Configure the service environment to use the secret file: `GEOSERVER_ADMIN_PASSWORD_FILE=/run/secrets/geoserver_admin_password`

### Versions

kartoza/geoserver:2.27.1

### Additional context

The issue lies in the logic of `update_passwords.sh`. The script checks if GEOSERVER_ADMIN_PASSWORD is empty and, if so, immediately generates a random string for it.

```python
# 1. Checks if env var is empty. It is empty (user used _FILE).
if [[ -z ${GEOSERVER_ADMIN_PASSWORD} ]]; then
      generate_random_string 15
      GEOSERVER_ADMIN_PASSWORD=${RAND} # <--- Variable is now set!
      # ...
fi
```

The function `file_env` then checks:
```python
if [ "${!var:-}" ] && [ "${!fileVar:-}" ]; then
    printf >&2 'error: both %s and %s are set (but are exclusive)\n' ...
    exit 1
fi
```

Since firt GEOSERVER_ADMIN_PASSWORD  is populated with a random string, and then found the _FILE variable, the  check fails and the password update is aborted.

Possible fix:
In `update_passwords.sh` the call to `file_env  GEOSERVER_ADMIN_PASSWORD `should be moved before the block that generates the random password. This ensures the script prefers the file secret if it exists, and only generates a random password if neither the environment variable nor the secret file are provided.

NyakudyaA:
@vchaparro can you check if the PR #748 update password if it resolves it

vchaparro:
Last version 2.28.2 doesn't fix the issue.

NyakudyaA:
[...]

PR #791, opened 2026-02-20
fix: correct update_passwords logic to allow Docker Secrets

vchaparro:
Fix issue #790


NyakudyaA (changes requested):
@vchaparro You mention password but i see you updated the env for USER, is this correct

vchaparro:
Hi,
I’ve created a new PR with the solution that worked for me. 
PR #803

PR #803, opened 2026-03-17
Fix issue #790

vchaparro:
This is what eventually worked for me to fix the issue #790


NyakudyaA:
@vchaparro can you fix the merge conflict so that we can merge

PR #818, opened 2026-05-31
run cron to update images

### LINKED ISSUE #816: Auto publish images
### Feature description

We need to be able to auto publish images once a new version becomes available. There are two modalities we need to cater for

- [ ] We are building with tomcat base images, we need logic to be able to push a new version if a new tomcat version becomes available. The challenge here is that we have a script to determine a new version of tomcat, do we update the base image or only dated version I.e 2.28.4 or 2.28.4-date images 
- [ ] The Geoserver new versions have a release cycle of 2 months but the bugfix not so, we can add a cron schedule to the GitHub action to check if the version upstream matches the published images and then publish them

### Additional context

I think the base image for tomcat updates are important for

- [ ] security reasons
- [ ] to get latest features 

For Geoserver version the check should be for 

- [ ] 3 months for stable version
- [ ] ideally weekly or
- [ ] monthly to check bugfix

How should we compile the following @tharanathkartoza

NyakudyaA:
fixes #816

### DIFF
--- a/.github/workflows/check-image.yaml
+++ b/.github/workflows/check-image.yaml
@@ -1,0 +1,111 @@
+name: daily-image-updates
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 2 * * *'
+
+
+
+jobs:
+  check-image:
+
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+
+    steps:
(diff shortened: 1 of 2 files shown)


NyakudyaA:
Ping @tharanathkartoza

